PR #1, opened 2018-01-10
Initial move from vertx-hawkular-metrics to global vertx-monitoring

jotak:
Based on PR https://github.com/vert-x3/vertx-hawkular-metrics/pull/65 from @tsegismont and @dankristensen

Includes only influxdb reporter at the moment

Issue #2, opened 2018-01-10
Application-defined metrics


jotak:
Have a standard way to manage application-defined metrics. Currently the hawkular implementation uses a dedicated eventbus endpoint to collect application-defined data points. Which, obviously, affects eventbus internal metrics.

Maybe a more direct wiring should be proposed.

jotak:
Fixed via micrometer registry access ; properly documented

PR #9, opened 2018-01-15
Initial work

jotak:
Replaces #1 

Initial move from vertx-hawkular-metrics to global vertx-monitoring

Based on PR vert-x3/vertx-hawkular-metrics#65 from @tsegismont and @dankristensen

Includes only influxdb reporter at the moment


jotak:
@vietj or @tsegismont can you review please?

vietj:
I've assigned Thomas formally to this, I'll throw an eye though to the PR.

batching-collector/pom.xml
```diff
@@ -26,7 +26,7 @@
 
   <artifactId>vertx-monitoring-batching-collector</artifactId>
 
-  <name>Vert.x Metrics Collector</name>
+  <name>Vert.x Monitoring Batching Collector</name>
```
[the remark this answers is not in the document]
tsegismont: In fact, why split the project into _common_ and _collector_?

There's just one collector implementation, it's not meant to be generic. In the original PR (https://github.com/vert-x3/vertx-hawkular-metrics/pull/65) there was just a _collector_ project

batching-collector/src/main/java/io/vertx/ext/monitoring/collector/impl/BatchingVertxMetrics.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.Map;
+
+import static io.vertx.ext.monitoring.common.MetricsCategory.*;
+
+/**
+ * Metrics SPI implementation.
+ *
+ * @author Thomas Segismont
+ */
+public abstract class BatchingVertxMetrics<T extends BatchingReporterOptions> extends DummyVertxMetrics {
```
tsegismont: See the other comment: in the original PR, this class was named `VertxMetricsBase`. It is not a _batching_ Vert.x Metrics impl. It's the base HTTP reporter class which is _batching_: it sends metrics in batches to the remote server.

batching-collector/src/main/java/io/vertx/ext/monitoring/collector/BatchingReporterOptions.java
```diff
@@ -26,27 +26,27 @@
...
  */
 @DataObject(generateConverter = true, inheritConverter = true)
-public abstract class BatchingReporterOptions extends MetricsOptionsBase {
+public class BatchingReporterOptions extends MetricsOptionsBase {
 
   /**
    * The default metric name prefix (empty).
    */
-  public static final String DEFAULT_PREFIX = "";
+  private static final String DEFAULT_PREFIX = "";
```
tsegismont: Default values are always exposed in Vert.x options classes

batching-collector/src/main/java/io/vertx/ext/monitoring/collector/impl/Scheduler.java
```diff
@@ -0,0 +1,89 @@
...
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Collects metrics and relay them to the sender.
+ *
+ * @author Thomas Segismont
+ */
+public class Scheduler {
+  private final Vertx vertx;
+  private final Handler<List<DataPoint>> sender;
```
tsegismont: Now that there's a `Reporter` interface, we should probably use this instead of the generic `Handler<List<DataPoint>>`, wdyt?

reporters/influxdb/README.adoc
```diff
@@ -0,0 +1,27 @@
...
+This project is an InfluxDb implementation of the Vert.x Metrics SPI: metrics built from Vert.x events will be sent to
+an InfluxDb Timeseries database.
+
+Please see the in-source Asciidoc documentation or the main documentation on the web-site for a full description
+of this component:
+
+* link:http://vertx.io/docs/vertx-monitoring-influxdb/[web-site docs]
+* link:src/main/asciidoc[in-source docs]
+
+If you want to test you can use this docker command to run an InfluxDb instance:
```
tsegismont: Could you rephrase this? It seems to imply you need this to run the Influx reporter test while we're actually mocking the Influx server

reporters/influxdb/src/test/java/io/vertx/ext/monitoring/influxdb/impl/InfluxDbReporterITest.java
```diff
@@ -0,0 +1,76 @@
...
+  public void shouldSendDataToInfluxDb(TestContext context) {
+    Async async = context.async();
+    vertx.createHttpServer(new HttpServerOptions()
+      .setCompressionSupported(true)
+      .setDecompressionSupported(true)
+      .setLogActivity(true)
+      .setHost("localhost")
+      .setPort(8086))
+      .requestHandler(req -> {
+        req.exceptionHandler(t -> System.out.println("An exception occured handling request: " + t));
```
tsegismont: Fail the `context`?

batching-collector/pom.xml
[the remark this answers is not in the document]
jotak: This is because I have a Prometheus reporter ready for a next PR, and it will be using common but not collector. So it's already prepared for split.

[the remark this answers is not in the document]
jotak: About why _batching_: 
In the previous PR HttpReporterBase was defined in collector and was setting down the mechanism for sending batches to reporters (it still does). So I wanted to make it clear that this "collector" module is actually designed to send batches to the reporter.

However I don't mind to change back to _collector_ to make it simpler.

batching-collector/src/main/java/io/vertx/ext/monitoring/collector/impl/BatchingVertxMetrics.java
jotak: `VertxMetricsBase` is too generic while this will only be used to send batches with a scheduler. Again, Prometheus won't use it. So I had to find something less generic than just _XXBase_.

vietj:
before going further, the package should be `io.vertx.monitoring` and not `io.vertx.ext.monitoring`

vietj:
the module collector module shall simply be named `vertx-monitoring-collector` (no need to say it's batching)

the module directories should be named:

- vertx-monitoring-collector
- vertx-monitoring-common
- vertx-monitoring-reporters

batching-collector/pom.xml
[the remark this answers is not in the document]
tsegismont: I would prefer to see all the code in this PR, to get a better picture. Is that ok @vietj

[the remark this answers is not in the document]
tsegismont: >In the previous PR HttpReporterBase was defined in collector and was setting down the mechanism for sending batches to reporters (it still does)

Yes, `HttpReporterBase`. Not `VertxMetricsBase`. So the collector is not a _batching_. It's only the base reporter class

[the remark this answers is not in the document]
vietj: I agree with @tsegismont I also wondered where the Prometheus support was. An initial-work PR should be complete.

batching-collector/src/main/java/io/vertx/ext/monitoring/collector/impl/BatchingVertxMetrics.java
tsegismont: I need to see the code for Prometheus. It sounds like we're going to maintain multiple collector implementations. We wanted to avoid that.

batching-collector/pom.xml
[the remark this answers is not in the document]
jotak: It was clear for me that in a first time I would take in just InfluxDB (that's in the PR head comment). I've never been aware than prometheus was expected for 3.5.1
Btw "completeness" depends on what you expect. It's 100% ready for use.

batching-collector/src/main/java/io/vertx/ext/monitoring/collector/impl/BatchingVertxMetrics.java
jotak: You can see it here [1], although it's not yet adapted to the new module "vertx-monitoring", it's roughly what I plan to put in.

[1] https://github.com/jotak/vertx-metrics/tree/prom/reporters/prometheus/src/main/java/io/vertx/ext/prometheus/impl

And yes it doesn't fit in this collector implementation. I talked about it previously (don't remember where and when but I can search). This is mainly because of the pull-based model. This is also because by using the prometheus registry (from the prometheus java client), we gain some functionalities that we don't have otherwise (collecting summaries, histograms, building timers out of summaries, etc.). I think there's too much to lose if we absolutely wanted to do something very generic.

jotak: Note also that some metrics colelcted for influx are pointless to collect for prometheus. Like eventbus' `receivedMessages` / `receivedLocalMessages` / `receivedRemoteMessages`. In prometheus it's just 1 metric with a label.

batching-collector/pom.xml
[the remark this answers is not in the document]
tsegismont: > It was clear for me that in a first time I would take in just InfluxDB (that's in the PR head comment)

I noticed your comment. I asked to include the Prometheus code as it seems we will not be reusing the collector implementation in the end.

>I've never been aware than prometheus was expected for 3.5.1

Me neither. But it seems to impact the way this project will work so I think it's better to consider what has been done already.

jotak:
Feedbacks done
I also added prometheus.

vertx-monitoring-reporters/vertx-monitoring-prometheus/README.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+* link:http://vertx.io/docs/vertx-monitoring-prometheus/[web-site docs]
+* link:src/main/asciidoc[in-source docs]
+
+== Quick setup
+
+To use the Prometheus reporter, you need first to get a running Prometheus instance.
+See link:https://prometheus.io/docs/prometheus/latest/installation/[Prometheus installation].
+
+Configure Prometheus to scrape your Vert.X server. Example in `prometheus.yml`:
```
tsegismont: `Vert.x` not `Vert.X`

vertx-monitoring-reporters/vertx-monitoring-prometheus/README.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+```yml
+scrape_configs:
+  - job_name: 'myapp'
+    static_configs:
+      - targets: ['localhost:8080']
```
tsegismont: Use another port as many users will have http servers running on this port

vertx-monitoring-reporters/vertx-monitoring-prometheus/pom.xml
```diff
@@ -0,0 +1,227 @@
...
+      <version>${prometheus-client.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-monitoring-common</artifactId>
+      <version>3.5.1-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-web</artifactId>
```
tsegismont: We cannot require to add Vert.x Web for monitoring with Prometheus

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/asciidoc/groovy/index.adoc
```diff
@@ -0,0 +1,361 @@
...
+
+----
+
+== Configuration
+
+=== Using an embedded HTTP server for exposing endpoints
+
+[source,groovy]
+----
+Code not translatable
```
tsegismont: Update the example so that it can be translated to Groovy

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/asciidoc/groovy/index.adoc
```diff
@@ -0,0 +1,361 @@
...
+----
+def vertx = Vertx.vertx([
+  metricsOptions:[
+    enabled:true
+  ]
+])
+
+// Later on, creating a router
+def router = Router.router(vertx)
+router.route("/custom").handler(io.vertx.monitoring.prometheus.impl.PrometheusVertxMetrics.createMetricsHandler())
```
tsegismont: It is nice to be able to plug a handler into an existing Vert.x Web app, avoiding to create bind multiple ports. But the reporter should be able to start a Vert.x core server if needed

vertx-monitoring-reporters/vertx-monitoring-prometheus/pom.xml
tsegismont: > I forgot to mark it optional. Would it be ok if it's optional?

Yes

> The reason for this is to allow "embedded prometheus server" to start automatically without the user having to manually create a webserver/router etc. for prometheus.

The reporter should start a Vert.x core HttpServer without requiring anything from the user else than some config.

>It's configurable in options, and if the user doesn't activate the embedded server then he doesn't have to provide that dependency.

How will the metrics get collected then?

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/asciidoc/groovy/index.adoc
jotak: It's the purpose of the above example, the one you cannot see because "code is not translatable" :)

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/asciidoc/groovy/index.adoc
```diff
@@ -0,0 +1,361 @@
...
+|===
+|Metric type
+|Metric name
+|Description
+
+|Gauge
+|`vertx_verticle{name=<name>}`
+|Number of verticle instances deployed.
+
+|===
```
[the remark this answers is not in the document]
jotak: Yeah it's a difference indeed, and it's somewhat related to issue #2 , basically at the moment the best way to have application-defined metrics with prometheus is to use the shared registry directly (just similar to the dropwizard impl).
I think this point could be addressed during the tech-preview phase, to make sure we provide a long term solution for this.

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/java/io/vertx/monitoring/prometheus/VertxPrometheusOptions.java
```diff
@@ -0,0 +1,140 @@
...
+  public VertxPrometheusOptions setEnabled(boolean enable) {
+    super.setEnabled(enable);
+    return this;
+  }
+
+  /**
+   * An embedded server will start to expose metrics with Prometheus format
+   * @param serverOptions the server options
+   */
+  public VertxPrometheusOptions embedServer(VertxPrometheusServerOptions serverOptions) {
```
tsegismont: Does this works well with codegen? Usually options classes have get/is,set methods.

tsegismont: It would be better to have an `enabled` field in `VertxPrometheusServerOptions`

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/java/io/vertx/monitoring/prometheus/VertxPrometheusServerOptions.java
```diff
@@ -0,0 +1,119 @@
...
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Vert.x Prometheus embedded server configuration.
+ *
+ * @author Joel Takvorian
+ */
+@DataObject(generateConverter = true, inheritConverter = true)
+public class VertxPrometheusServerOptions {
```
tsegismont: Can you add an `HttpServerOptions` field? Users could then configure the endpoint (ssl in particular)

vertx-monitoring-reporters/vertx-monitoring-prometheus/pom.xml
jotak: > The reporter should start a Vert.x core HttpServer without requiring anything from the user else than some config.

It's the case

> How will the metrics get collected then?

Without embedded server, the metrics are still collected in the registry, and that registry is accessible to the user so that he can do whatever he wants. It's interesting in the case where you want to reuse an existing router, as shown in one of the examples.

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/java/io/vertx/monitoring/prometheus/VertxPrometheusOptions.java
jotak: I'm changing this, maybe it's the reason why the groovy example is not generated

About the `enabled` field, I've seen it's used in a lot of option classes, but to be honest I'm a bit perplexed about it. If I set Options XYZ, why would I want to have it disabled?
In other words I would prefer to write: `new VertxPrometheusOptions().setEmbeddedServerOptions(new EmbeddedServerOptions());`, where it's implictly enabled, rather than 
`new VertxPrometheusOptions().setEmbeddedServerOptions(new EmbeddedServerOptions().setEnabled());` which adds verbosity for nothing.

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/java/io/vertx/monitoring/prometheus/VertxPrometheusServerOptions.java
jotak: Actually, I'm replacing `VertxPrometheusServerOptions` with `HttpServerOptions`.
I'm adding the `String endpoint` field directly in `VertxPrometheusOptions`

Doing so is much easier to use, no need to encapsulate options in options in options.

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/asciidoc/dataobjects.adoc
```diff
@@ -36,24 +44,12 @@ Set whether metrics will be enabled on the Vert.x instance. Metrics are not enab
...
-[frame="topbot"]
-|===
-^|Name | Type ^| Description
-|[[endpoint]]`endpoint`|`String`|-
-|[[host]]`host`|`String`|-
-|[[port]]`port`|`Number (int)`|-
+|[[registryName]]`registryName`|`String`|
++++
+Set a name for the prometheus registry, so that a new registry will be created and associated with this name.
+ To retrieve this registry later, call link
```
jotak: It seems that doc generation doesn't like all links...

vertx-monitoring-reporters/vertx-monitoring-influxdb/src/main/java/io/vertx/monitoring/influxdb/package-info.java
```diff
@@ -0,0 +1,424 @@
...
+ * If the metric is a counter or availability, or if you prefer explicit configuration, set the {@code type} and/or {@code timestamp} attributes:
+ *
+ * [source,$lang]
+ * ----
+ * {@link examples.MetricsExamples#userDefinedMetricExplicit()}
+ * ----
+ *
+ * NOTE: InfluxDb understands all timestamps as milliseconds since January 1, 1970, 00:00:00 UTC.
+ */
+@ModuleGen(name = "vertx-influxdb", groupPackage = "io.vertx")
```
jotak: @tsegismont I'm not sure if "ModuleGen" must take name of the artifact, or if it's not necessary?
Here the maven artifact is vertx-monitoring-influxdb

jotak:
Did some small updates as requested by @vietj (upgrade parent pom...). Also fixed doc gen for "links" in dataobjects.adoc.

batching-collector/src/main/java/io/vertx/ext/monitoring/collector/impl/BatchingVertxMetrics.java
tsegismont: @jotak yes I remember we talked about the problem but not about a solution. Or maybe you did with @vietj ? The thing is that having if we have two collector and reporters then there's no point in creating this module with reporter and collectors subprojects.

vertx-monitoring-reporters/vertx-monitoring-prometheus/pom.xml
tsegismont: Thanks for the details

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/asciidoc/groovy/index.adoc
tsegismont: Can you investigate why it's not translatable?

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/asciidoc/groovy/index.adoc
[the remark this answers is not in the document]
tsegismont: Is the event bus bridged used by the Prometheus reporter? It seems it is as there related fields in the options class.

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/java/io/vertx/monitoring/prometheus/VertxPrometheusOptions.java
tsegismont: Because Vert.x is polyglot and when you write JS or Ruby verticles, you don't even have option classes. Just JSON or Hashes.

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/asciidoc/groovy/index.adoc
jotak: It's fixed now. I had to use only pojo-style getters/setters in the examples.

batching-collector/src/main/java/io/vertx/ext/monitoring/collector/impl/BatchingVertxMetrics.java
jotak: Maybe @vietj can comment.
My understanding is that vertx-monitoring-parent will be the new parent for all metrics-related modules. For instance, dropwizard module could be move here later (but if we want to make it consistent we would have to rename it to smthg like vertx-monitoring-dropwizard, which is a breaking change, so for the time being we don't touch it)

But still, it's true that since hawkular is removed this architecture is now a bit weird with a generic-style collector that is used only for 1 reporter, influx. However it can be used later if a new push-based TSDB needs to be implemented here. So we may not want to destroy something that has been designed to handle various reporters. It's not as if we had a huge benefit of breaking it. My opinion is that we can keep it as is.

vertx-monitoring-reporters/vertx-monitoring-prometheus/src/main/asciidoc/groovy/index.adoc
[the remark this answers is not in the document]
jotak: Removed from prometheus options

tsegismont:
@jotak it won't be possible to resolve the duplication issue for 3.5.1. However, we want to release both reporters as a tech preview.

To move things forward, could you please move everything to a single `vertx-monitoring` Maven module ? This will avoid having many jars which may disappear in the next release. The Influx and Prometheus specific code can still be separated in different packages (two `package-info` files will produce two Asciidoc files). We'll then arrange the separation on the Vert.x website.

jotak:
Just merged everything into a single module. The doc files are incorrect though (still need to be sorted out)

src/main/java/examples/MetricsExamples.java
```diff
@@ -0,0 +1,182 @@
...
+import io.vertx.monitoring.influxdb.VertxInfluxDbOptions;
+import io.vertx.monitoring.prometheus.VertxPrometheusOptions;
+import io.vertx.monitoring.prometheus.impl.PrometheusVertxMetrics;
+
+/**
+ * @author Thomas Segismont
+ * @author Joel Takvorian
+ */
+@Source
+@SuppressWarnings("unused")
```
tsegismont: Please remove this. You shouldn't have unused methods in this class, they should all be referred from the Influx reporter `package-info` file.

Note that some examples look like they were copied from the Hawkular ones but are not relevant to Influx

src/main/asciidoc/js/influxdb.adoc
```diff
@@ -0,0 +1,419 @@
...
+|===
+
+== User defined metrics
+
+Users can send their own metrics to the InfluxDb server. In order to do so, the event bus metrics bridge must be
+enabled:
+
+[source,js]
+----
+Code not translatable
```
tsegismont: Some examples are still not translated

src/main/java/io/vertx/monitoring/VertxMetricsFactoryImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+      return new InfluxDbVertxMetrics(vertx, new VertxInfluxDbOptions(metricsOptions.toJson()));
+    }
+  }
+
+  @Override
+  public MetricsOptions newOptions() {
+    // Tricky hack to guess if we're expecting Prometheus options or Influx
+    try {
+      Class.forName("io.prometheus.client.CollectorRegistry");
```
tsegismont: Can you replace with:

```java
Class.forName("io.prometheus.client.CollectorRegistry", false, getClass().getClassLoader())
```

Then the class will not be initialized until it's actually used (useful if metrics are disabled).

src/main/java/io/vertx/monitoring/prometheus/impl/PrometheusVertxMetrics.java
```diff
@@ -0,0 +1,209 @@
...
+import static io.vertx.monitoring.common.MetricsCategory.*;
+
+/**
+ * Metrics SPI implementation for Prometheus.
+ *
+ * @author Joel Takvorian
+ */
+public class PrometheusVertxMetrics extends DummyVertxMetrics {
+  private final CollectorRegistry registry;
+  private final Optional<EventBusMetrics> eventBusMetrics;
```
tsegismont: Don't use `Optional` as a field. Its value is only used inside this class.

src/main/java/io/vertx/monitoring/prometheus/impl/PrometheusVertxMetrics.java
```diff
@@ -0,0 +1,209 @@
...
+   * Create a metrics handler for vertx router, using the registry identified by {@code registryName}
+   */
+  public static Handler<RoutingContext> createMetricsHandler(String registryName) {
+    // Create handler from io.prometheus:simpleclient_vertx
+    return new MetricsHandler(PrometheusRegistries.get(registryName));
+  }
+
+  private HttpServer startServer() {
+    HttpServerOptions serverOptions = options.getEmbeddedServerOptions();
+    Router router = Router.router(vertx);
```
tsegismont: Vert.x Web is not really optional if it's needed even for the embedded server. Or am I missing something?

src/main/java/io/vertx/monitoring/prometheus/package-info.java
```diff
@@ -0,0 +1,388 @@
...
+ * {@link io.vertx.core.eventbus.EventBus} and handlers
+ * * User defined metrics can be added to the shared registry (using Prometheus https://github.com/prometheus/client_java[simple Java client])
+ *
+ * == Prerequisites
+ *
+ * Follow the https://prometheus.io/docs/prometheus/latest/getting_started/[instructions to get Prometheus up and running].
+ *
+ * == Getting started
+ *
+ * The _${maven.artifactId}_ module must be present in the classpath.
```
tsegismont: Can you also say that the Prometheus client must be on the classpath?

tsegismont: BTW, why do we need this `simpleclient` JAR if we only expose metrics on an HTTP server?

src/main/java/io/vertx/monitoring/prometheus/VertxPrometheusOptions.java
```diff
@@ -0,0 +1,160 @@
...
+   * By default, disables <i>remote</i> label for net/http servers.
+   */
+  public static final boolean DEFAULT_ENABLE_REMOTE_LABEL_FOR_SERVERS = false;
+
+  /**
+   * The default metrics endpoint = /metrics when using an embedded server.
+   */
+  public static final String DEFAULT_EMBEDDED_SERVER_ENDPOINT = "/metrics";
+
+  private HttpServerOptions embeddedServerOptions;
```
tsegismont: Don't use null as an "enabled" indicator. This is not how we deal with options in general.

This field should be initialized with the default `HttpServerOptions` and then you can add an `startEmbeddedServer` boolean.

src/main/java/io/vertx/monitoring/prometheus/VertxPrometheusOptions.java
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * By default, uses the standard prometheus registry (name is null)
+   */
+  public static final String DEFAULT_REGISTRY_NAME = null;
+
+  /**
+   * By default, enables <i>remote</i> label for net/http clients.
+   */
+  public static final boolean DEFAULT_ENABLE_REMOTE_LABEL_FOR_CLIENTS = true;
```
tsegismont: 👍

src/main/java/io/vertx/monitoring/prometheus/VertxPrometheusOptions.java
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * By default, disables <i>remote</i> label for net/http servers.
+   */
+  public static final boolean DEFAULT_ENABLE_REMOTE_LABEL_FOR_SERVERS = false;
```
tsegismont: 👍

src/main/java/io/vertx/monitoring/prometheus/impl/Labels.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package io.vertx.monitoring.prometheus.impl;
+
+import io.vertx.core.net.SocketAddress;
+
+/**
+ * @author Joel Takvorian
+ */
+public final class Labels {
```
tsegismont: Why not an enum?

PR #12, opened 2018-01-31
Micrometer implementation of Vert.x metrics SPI

jotak:
This includes:

- InfluxDB reporting
- Prometheus reporting


src/main/java/io/vertx/monitoring/MetricsCategory.java
```diff
@@ -0,0 +1,59 @@
...
+package io.vertx.monitoring;
+
+import io.vertx.codegen.annotations.VertxGen;
+
+
+/**
+ * Metrics types for each metrics.
+ */
+@VertxGen
+public enum MetricsCategory {
```
jotak: This is sometimes referred as "metrics categories", sometimes as "domain". I should unify this... so maybe rename it "MetricsDomain"

src/main/java/io/vertx/monitoring/VertxMonitoringOptions.java
```diff
@@ -0,0 +1,234 @@
...
+    .setDomain(MetricsCategory.NET_SERVER)
+    .setLabel("remote")
+    .setType(MatchType.REGEX)
+    .setValue(".*")
+    .setAlias("_");
+
+  /**
+   * The default label matches: empty by default
+   */
+  public static final List<Match> DEFAULT_LABEL_MATCHES = Arrays.asList(DEFAULT_HTTP_SERVER_MATCH, DEFAULT_NET_SERVER_MATCH);
```
jotak: Not sure yet what is the best way to have default matches.
The idea is that by default the server "remote" label should be disabled to prevent too high cardinality on this label.
But that means that in order to enable it, the user has to call "options.resetLabelMatchs". If keeping like that, it should be better documented.

jotak:
@tsegismont I think it's mature for a review. I've added JMX reporting.

bckfnn:
I think it is great.

Do your label matching code add anything that micrometer can't do with MeterFilters? I would guess it is mostly about configuration from json?

I would suggest that some default matches are added, like DEFAULT_HTTP_SERVER_MATCH, which also reduce eventbus "address" tag, httpserver "path" and httpclient "path" and "remote". Perhaps not included in DEFAULT_LABEL_MATCHES but as examples of potential high cardinality series.

jotak:
@bckfnn I'm doing some tests with micrometer's `MeterFilter`. That could be interesting indeed and we would avoid creating a specific api for this.

tsegismont:
@jotak make sure that the resulting code will be polyglot (a JS verticle developer can still define a set of filters)

jotak:
Yes, there are several ways to implement it.

One possibility is to throw away the generic "label matchers" that I did and put back the specific matchers for eb handlers / client uris / server uris as they are in the dropwizard module. And aside from this, document the possibility to directly set micrometer `MeterFilter` on the registry, since it's accessible. But that would mean configuration is done from code and not from config object / json.

Another possibility is still to document how to create accurate filters on micrometer registry, but in addition provide a "best-effort" mapping to create them out of a config object.

jotak:
So I kept the "label matchers" but removed the algorithmic parts to rely on MeterFilters. It simplifies our code with no loss in capabilities.

jotak:
Replacing with PR #14 (same content, but coming from vertx repo instead of my own)

PR #14, opened 2018-02-27
Micrometer implementation of Vert.x metrics SPI

jotak:
This includes:

- InfluxDB reporting
- Prometheus reporting
- JMX reporting

This is the same as PR #12 , recreated from another branch (from vert-x3 repo instead of my own)


bckfnn:
It is great to see this feature moving forward.
I notices that there is a test dependency on slf4j version 1.8. Micrometer 1.0.0 no longer have a dependency on 1.8 but uses the 1.7 version instead.

jotak:
Thanks for the tip @bckfnn , it's updated.

README.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+Vert.x Micrometer Metrics contains metrics collection and reporting to various target systems through link:http://micrometer.io/[Micrometer].
+
+The http://vertx.io/docs/vertx-core/java/index.html#_metrics_spi[Vert.x Metrics SPI] allows implementers to
+capture events from Vert.x in order to gather metrics.
+
+Please see the in-source Asciidoc documentation or the main documentation on the web-site for a full description
+of this component:
+
+* link:http://vertx.io/docs/vertx-micrometer-metrics/[web-site docs]
```
tsegismont: Shouldn't this point to `http://vertx.io/docs/vertx-micrometer-metrics/java` ?

README.adoc
```diff
@@ -0,0 +1,61 @@
...
+* link:src/main/asciidoc[in-source docs]
```
tsegismont: `/src/main/asciidoc/java/index.adoc`

README.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+== InfluxDB
+
```
tsegismont: Remove this section

README.adoc
```diff
@@ -0,0 +1,61 @@
...
+docker run --rm --name vertx-influxdb -p 8086:8086 -p 8083:8083 -e INFLUXDB_ADMIN_ENABLED=true influxdb
+```
+
+You can connect to the administration interface in your browser at `http://localhost:8083`.
+You need to create a database from this interface. By default, the reporter will use `default`, but you can change it
+through `VertxInfluxDbOptions#setDatabase`.
+
+InfluxDB will listen for api requests on port 8086 by default.
+
+== Prometheus
```
tsegismont: Remove this section

pom.xml
```diff
@@ -0,0 +1,275 @@
...
+  ~  You may elect to redistribute this code under either of these licenses.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.vertx</groupId>
+    <artifactId>vertx-ext-parent</artifactId>
+    <version>28</version>
```
tsegismont: Have you checked this is the right version for `3.5.1`?

pom.xml
```diff
@@ -0,0 +1,275 @@
...
+              <goal>verify</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+
+  <profiles>
+    <profile>
```
tsegismont: why this empty profile?

src/main/java/io/vertx/micrometer/package-info.java
```diff
@@ -0,0 +1,594 @@
...
+ * = Vert.x Micrometer Metrics
+ *
+ * This project is an implementation of the Vert.x Metrics Service Provider Interface (SPI).
+ * It uses link:http://micrometer.io/[Micrometer] for managing metrics and reporting to several backends.
+ *
+ * == Features
+ *
+ * * Vert.x core tools monitoring: TCP/HTTP client and servers, {@link io.vertx.core.datagram.DatagramSocket},
+ * {@link io.vertx.core.eventbus.EventBus} and pools
+ * * User defined metrics through micrometer
```
tsegismont: `Micrometer`

src/main/java/io/vertx/micrometer/package-info.java
```diff
@@ -0,0 +1,594 @@
...
+ * * Reporting to https://www.influxdata.com/[InfluxDB], https://prometheus.io/[Prometheus] or JMX.
```
tsegismont: Micrometer supports other backends. If I want to report to Graphite or Cloudwatch, how can I do that?

src/main/java/io/vertx/micrometer/package-info.java
```diff
@@ -0,0 +1,594 @@
...
+ *
+ * == InfluxDB
+ *
+ * === Prerequisites
+ *
+ * Follow the https://docs.influxdata.com/influxdb/latest/introduction/getting_started/[instructions to get InfluxDb up and running].
+ *
+ * NOTE: You can use a standalone InfluxDB server as well.
```
tsegismont: It seems to me this note was copied over from the original Hawkular implementation

src/main/java/io/vertx/micrometer/package-info.java
```diff
@@ -0,0 +1,594 @@
...
+ * ----
+ * <dependency>
+ *   <groupId>${maven.groupId}</groupId>
+ *   <artifactId>${maven.artifactId}</artifactId>
+ *   <version>${maven.version}</version>
+ * </dependency>
+ * <dependency>
+ *   <groupId>io.micrometer</groupId>
+ *   <artifactId>micrometer-registry-jmx</artifactId>
+ *   <version>1.0.0-rc.9</version>
```
tsegismont: In this doc, you can use Maven properties instead of hardcoded versions (with some POM config)

src/main/java/io/vertx/micrometer/backend/VertxJmxMetricsOptions.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  public boolean isEnabled() {
+    return enabled;
+  }
+
+  /**
+   * Set true to enable Prometheus reporting
+   */
+  public VertxJmxMetricsOptions setEnabled(boolean enabled) {
```
tsegismont: Why do we need this?

src/main/java/io/vertx/micrometer/backend/VertxPrometheusOptions.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  public boolean isEnabled() {
+    return enabled;
+  }
+
+  /**
+   * Set true to enable Prometheus reporting
+   */
+  public VertxPrometheusOptions setEnabled(boolean enabled) {
```
tsegismont: Why do we need this?

src/main/java/io/vertx/micrometer/backend/VertxInfluxDbOptions.java
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  public boolean isEnabled() {
+    return enabled;
+  }
+
+  /**
+   * Set true to enable InfluxDB reporting
+   */
+  public VertxInfluxDbOptions setEnabled(boolean enabled) {
```
tsegismont: Why do we need this?

src/main/java/io/vertx/micrometer/backend/BackendRegistries.java
```diff
@@ -0,0 +1,175 @@
...
+  /**
+   * Create a new backend registry, containing a micrometer registry, initialized with the provided options.
+   * If a registry already exists with the associated name, it is just returned without any effect.
+   * @param vertx vertx object that might be necessary for some backend initialization
+   * @param options micrometer options, including configuration related to the backend.
+   *                Should be a subclass of {@link MicrometerMetricsOptions} (ex: {@link VertxInfluxDbOptions}, {@link VertxPrometheusOptions}).
+   *                If the class is not recognized, a {@link NoopBackendRegistry} will be returned.
+   * @return the created (or existing) {@link BackendRegistry}
+   */
+  public static BackendRegistry setupBackend(Vertx vertx, MicrometerMetricsOptions options) {
```
tsegismont: So we can have only one registry at a time? The limitation is on our side correct? In this case, why?

src/main/java/io/vertx/micrometer/package-info.java
```diff
@@ -0,0 +1,594 @@
...
+ * Please refer to {@link io.vertx.micrometer.MicrometerMetricsOptions} for an exhaustive list of options.
+ *
+ * === Disable some metric domains
+ *
+ * Restricting the Vert.x modules being monitored can be done using
+ * {@link io.vertx.micrometer.MicrometerMetricsOptions#disabledMetricsCategories}.
+ *
+ * For a full list of domains, see {@link io.vertx.micrometer.MetricsDomain}
+ *
+ * === User-defined metrics
```
tsegismont: Would be nice to have an example of user defined metric and usage. Like a timer on an async method.

README.adoc
jotak: I don't know, do we want to put Java first? Or leave the choice of the language to the user?

README.adoc
jotak: I don't mind removing, but last time when I added the Prometheus section I got the comment "hey you should write a quick startup guide as we have for influx", so, I just want to make sure we're all ok to remove and won't wonder later why the quick startup guide disappeared :)

pom.xml
jotak: What's the correct way to check that? Just looking at another module, it also points to 28 for 3.5.1 : https://github.com/vert-x3/vertx-jdbc-client/commit/09d82385b3eebc5e49e32d87df6a119ced7db378

src/main/java/io/vertx/micrometer/package-info.java
jotak: It's not possible at the moment. Each backend requires the create its own registry implementation in Micrometer (ex: `GraphiteMeterRegistry`), with its own options (`GraphiteConfig`) and we would have to support that in relation with our own VertxOptions.

It could be possible to provide a alternate way, leaving to the user the possibility to create any registry of his own and inject it in `VertxMetricsImpl` ... at the moment this is not possible and I'm not sure we want to allow such different ways of configuring metrics.

Note that implementing a new backend (that is supported by micrometer) doesn't require a lot of effort so we could easily add more in the future.

pom.xml
jotak: removing... old module legacy

src/main/java/io/vertx/micrometer/package-info.java
jotak: removing

src/main/java/io/vertx/micrometer/package-info.java
jotak: That would be nice indeed... Moreover the version is incorrect here

src/main/java/io/vertx/micrometer/backend/VertxJmxMetricsOptions.java
jotak: Errr, what did I misunderstand here? In a previous comment you were asking me to add `enabled` flags to such embedded option objects [1] due to the polyglot nature of Vert.x (iiuc we cannot rely on null/absent objects depending on the language). Why wouldn't it be the case here?

[1] https://github.com/vert-x3/vertx-micrometer-metrics/pull/9#discussion_r162611481

src/main/java/io/vertx/micrometer/backend/BackendRegistries.java
jotak: True. We could change all the options to allow any number of backends of any types but I've never heard it as a requirement, so at some point we have to delimit the scope of what we want.

src/main/java/io/vertx/micrometer/package-info.java
jotak: Will do that

README.adoc
jotak: Actually it would currently lead to a 404 error, so I'll change to Java

src/main/java/io/vertx/micrometer/package-info.java
jotak: done

README.adoc
jotak: done

README.adoc
jotak: done

README.adoc
jotak: done

README.adoc
jotak: removed

pom.xml
tsegismont: Then we should be fine

src/main/java/io/vertx/micrometer/package-info.java
tsegismont: OK. Let's not block the release with this. But in the end, the user should be able to report to any existing micrometer backend without submitting a PR and waiting for a release. Can you file an issue? We will address it in the tech preview phase.

src/main/java/io/vertx/micrometer/backend/VertxJmxMetricsOptions.java
tsegismont: Right :+1:

src/main/java/io/vertx/micrometer/backend/BackendRegistries.java
tsegismont: No pb. Again, let's not block the release for this but can you file an issue that we'll address during tech preview phase? Thanks.

tsegismont:
@jotak can you reorganize as suggested in the last review comment (interfaces and data classes in the same root package, implementation in `impl` package?)

tsegismont:
@jotak trying to run tests, I got:

```
Running io.vertx.micrometer.backend.PrometheusMetricsITest
[vert.x-eventloop-thread-1] 2018-03-16T23:03:41.187+01:00 SEVERE [io.vertx.core.http.impl.HttpClientRequestImpl]  io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:9090
Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 13.895 sec <<< FAILURE! - in io.vertx.micrometer.backend.PrometheusMetricsITest
shouldStartEmbeddedServer(io.vertx.micrometer.backend.PrometheusMetricsITest)  Time elapsed: 10.025 sec  <<< ERROR!
java.util.concurrent.TimeoutException: Timed out
	at io.vertx.ext.unit.impl.CompletionImpl.awaitSuccess(CompletionImpl.java:99)
	at io.vertx.micrometer.backend.PrometheusMetricsITest.shouldStartEmbeddedServer(PrometheusMetricsITest.java:69)
```

jotak:
Hmm ... I could expect some timing failures, especially if run on jenkins CI, but not on this test where it just starts the server.

I just checked again on my laptop it runs well.

I'm trying right now on the CI server, I may get some timing issues to fix.
https://vertx.ci.cloudbees.com/job/vert.x3-monitoring/

jotak:
Actually, even jenkins is happy! https://vertx.ci.cloudbees.com/job/vert.x3-monitoring/97/console

tsegismont:
Just tried 4 times, got 2 successes, the same error again (`Connection refused: localhost/127.0.0.1:9090`) and this

```
Failed tests: 
  VertxNetClientServerMetricsTest.shouldReportHttpServerMetrics:95 
Expecting:
  <[vertx.net.server.bytesReceived[local=localhost:9194,remote=_]$COUNT/204.0,
    vertx.net.server.bytesReceived[local=localhost:9194,remote=_]$TOTAL/2448.0,
    vertx.net.server.connections[local=localhost:9194,remote=_]$VALUE/2.0,
    vertx.net.server.bytesSent[local=localhost:9194,remote=_]$COUNT/204.0,
    vertx.net.server.bytesSent[local=localhost:9194,remote=_]$TOTAL/1836.0]>
to contain only:
  <[vertx.net.server.connections[local=localhost:9194,remote=_]$VALUE/0.0,
    vertx.net.server.bytesReceived[local=localhost:9194,remote=_]$COUNT/204.0,
    vertx.net.server.bytesReceived[local=localhost:9194,remote=_]$TOTAL/2448.0,
    vertx.net.server.bytesSent[local=localhost:9194,remote=_]$COUNT/204.0,
    vertx.net.server.bytesSent[local=localhost:9194,remote=_]$TOTAL/1836.0]>
elements not found:
  <[vertx.net.server.connections[local=localhost:9194,remote=_]$VALUE/0.0]>
and elements not expected:
  <[vertx.net.server.connections[local=localhost:9194,remote=_]$VALUE/2.0]>
```

jotak:
Damn, I've run `mvn install` or `mvn clean install` 8 times in a row without any failure :/
What could I be missing? Do you have some option to run tests in parallel, or something like that?
Ports are hardcoded in my tests. Could it be the issue?

tsegismont:
No, I don't run tests in parallel. I don't think the hardcoded ports are an
issue. Can you try to annotate the tests with @Repeat and see if you can
get the failure as well?

2018-03-20 8:57 GMT+01:00 Joel Takvorian <notifications@github.com>:

src/main/java/io/vertx/micrometer/backends/PrometheusBackendRegistry.java
```diff
@@ -0,0 +1,74 @@
...
+      HttpServerOptions serverOptions = options.getEmbeddedServerOptions();
+      if (serverOptions == null) {
+        serverOptions = new HttpServerOptions();
+      }
+      Router router = Router.router(vertx);
+      router.route(options.getEmbeddedServerEndpoint()).handler(routingContext -> {
+        String response = registry.scrape();
+        routingContext.response().end(response);
+      });
+      server = vertx.createHttpServer(serverOptions)
```
jotak: @tsegismont , looking for an advice here: I don't set any exception handler here (this is where we start the prometheus server), does it mean any exception will be silently discarded (which mean we should rather log it) or is it logged by default?

jotak: @tsegismont
I could reproduce some issues when setting `@Repeat(50)`, so I've put in place a retry mechanism and it seems ok now.

jotak:
@tsegismont
I could reproduce some issues when setting `@Repeat(50)`, so I've put in place a retry mechanism and it seems ok now.

src/main/java/io/vertx/micrometer/backends/PrometheusBackendRegistry.java
tsegismont: You can only get an exception when you call `listen`. Or do I miss something?

tsegismont:
@jotak thanks for the update, now the tests always pass on my machine

src/main/java/io/vertx/micrometer/VertxJmxMetricsOptions.java
```diff
@@ -0,0 +1,132 @@
...
+
+import java.time.Duration;
+
+/**
+ * Options for Prometheus metrics backend.
+ *
+ * @author Joel Takvorian
+ */
+@DataObject(generateConverter = true, inheritConverter = true)
+public class VertxJmxMetricsOptions implements JmxConfig {
```
tsegismont: @jotak why do we need to extend JmxConfig?

tsegismont: Same goes for InfluxDb options. I'm wondering if that could cause any problem with codegen.

src/main/java/io/vertx/micrometer/VertxJmxMetricsOptions.java
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Set true to enable Prometheus reporting
+   */
+  public VertxJmxMetricsOptions setEnabled(boolean enabled) {
+    this.enabled = enabled;
+    return this;
+  }
+
+  public String getDomain() {
```
tsegismont: Can you add Javadoc? Here and on other getters in all option classes?

src/main/java/io/vertx/micrometer/VertxJmxMetricsOptions.java
tsegismont: Ha I see. It's used to configure the Micrometer registry. The data object fields correspond to methods in the config interface, correct?

src/main/java/io/vertx/micrometer/MetricsService.java
```diff
@@ -0,0 +1,76 @@
...
+
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.metrics.Measured;
+import io.vertx.micrometer.impl.MetricsServiceImpl;
+
+import java.util.Set;
+
+/**
+ * The metrics service mainly allows to return a snapshot of measured objects.
```
tsegismont: Can you add a note indicating this was taken and adapted from the `vertx-dropwizard-metrics` projects?

src/main/java/io/vertx/micrometer/MetricsService.java
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * @return the known metrics names by this service
+   */
+  Set<String> metricsNames();
+
+  /**
+   * Will return the metrics that correspond with the {@code measured} object, null if no metrics is available.<p/>
+   *
+   * Note: in the case of scaled servers, the JsonObject returns an aggregation of the metrics as the
+   * dropwizard backend reports to a single server.
```
tsegismont: Still mentions Dropwizard. Is the comment still valid here?

src/main/java/io/vertx/micrometer/MetricsService.java
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * @return the map of metrics where the key is the name of the metric (excluding the base name unless for the Vert.x object)
+   * and the value is the json data representing that metric
+   */
+  JsonObject getMetricsSnapshot();
+
+  /**
+   * Will return the metrics that begins with the {@code baseName}, null if no metrics is available.<p/>
+   *
+   * Note: in the case of scaled servers, the JsonObject returns an aggregation of the metrics as the
```
tsegismont: Still mentions Dropwizard. Is the comment still valid here?

src/main/java/io/vertx/micrometer/MicrometerMetricsOptions.java
```diff
@@ -0,0 +1,248 @@
...
+@DataObject(generateConverter = true, inheritConverter = true)
+public class MicrometerMetricsOptions extends MetricsOptions {
+
+  /**
+   * Default registry name is 'default'
+   */
+  public static final String DEFAULT_REGISTRY_NAME = "default";
+
+  /**
+   * Default label match for public http server: exclude remote label
```
tsegismont: Perhaps there should be in the note in the module documentation as well.

src/main/java/io/vertx/micrometer/MicrometerMetricsOptions.java
```diff
@@ -0,0 +1,248 @@
...
+   */
+  public static final Match DEFAULT_HTTP_SERVER_MATCH = new Match()
+    .setDomain(MetricsDomain.HTTP_SERVER)
+    .setLabel("remote")
+    .setType(MatchType.REGEX)
+    .setValue(".*")
+    .setAlias("_");
+
+  /**
+   * Default label match for public net server: exclude remote label
```
tsegismont: Perhaps there should be in the note in the module documentation as well.

src/main/java/io/vertx/micrometer/backends/PrometheusBackendRegistry.java
```diff
@@ -0,0 +1,74 @@
...
+        .requestHandler(router::accept)
+        .listen(serverOptions.getPort(), serverOptions.getHost());
```
tsegismont: @jotak can you add a handler and if the server can't be started log the problem ?

src/main/java/io/vertx/micrometer/impl/VertxHttpClientMetrics.java
```diff
@@ -0,0 +1,152 @@
...
+      return null;
+    }
+
+    @Override
+    public void closeEndpoint(String host, int port, Void endpointMetric) {
+    }
+
+    @Override
+    public Void enqueueRequest(Void endpointMetric) {
+      return null;
```
tsegismont: Can you file an issue so that we later implement endpoint metrics as in `vertx-dropwizard-metrics`?

src/main/java/io/vertx/micrometer/impl/AbstractMetrics.java
```diff
@@ -0,0 +1,75 @@
...
+
+  Counters counters(String name, String description, Label... keys) {
+    return new Counters(domain.getPrefix() + name, description, registry, keys);
+  }
+
+  Gauges<LongAdder> longGauges(String name, String description, Label... keys) {
+    return new Gauges<>(domain.getPrefix() + name, description, LongAdder::new, LongAdder::doubleValue, registry, keys);
+  }
+
+  Gauges<AtomicReference<Double>> doubleGauges(String name, String description, Label... keys) {
```
tsegismont: How about using a `java.util.concurrent.atomic.DoubleAdder` here? With the same benefits as in `longGauges`?

src/main/java/io/vertx/micrometer/impl/Labels.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    List<Tag> tags = new ArrayList<>(keys.length);
+    for (int i = 0; i < keys.length; i++) {
+      String lowKey = keys[i].toString().toLowerCase();
+      tags.add(Tag.of(lowKey, values[i]));
+    }
+    return tags;
+  }
+
+  public static class Values {
```
tsegismont: If the enclosing class is `final`, I suppose you want to make this one `final` too.

src/main/java/io/vertx/micrometer/impl/VertxMetricsImpl.java
```diff
@@ -0,0 +1,173 @@
...
+    verticleMetrics.ifPresent(vm -> vm.verticleDeployed(verticle));
+  }
+
+  @Override
+  public void verticleUndeployed(Verticle verticle) {
+    verticleMetrics.ifPresent(vm -> vm.verticleUndeployed(verticle));
+  }
+
+  @Override
+  public void timerCreated(long l) {
```
tsegismont: Can you file an issue so that we implement timer metrics later?

src/main/java/io/vertx/micrometer/impl/VertxMetricsImpl.java
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Metrics SPI implementation for Micrometer.
+ *
+ * @author Joel Takvorian
+ */
+public class VertxMetricsImpl extends AbstractMetrics implements VertxMetrics {
+  private final BackendRegistry backendRegistry;
+  private final String registryName;
+  private final Optional<EventBusMetrics> eventBusMetrics;
```
tsegismont: We should avoid `Optional` here. These are not exposed at all, and if `Optional` is convenient in some cases for return types, it shouldn't be used for a field.

tsegismont:
@jotak I'm done with the implementation review. After the remaining concerns are addressed, I'll merge this one. Good job!

src/main/java/io/vertx/micrometer/VertxJmxMetricsOptions.java
jotak: Yes, that's it. In other words it avoids to create a conversion method. I think it makes sense.

jotak: Actually, I will create a converter. It's less ambiguous

src/main/java/io/vertx/micrometer/MetricsService.java
jotak: Not valid anymore, removing...

src/main/java/io/vertx/micrometer/VertxJmxMetricsOptions.java
tsegismont: :+1:

src/main/java/io/vertx/micrometer/backends/PrometheusBackendRegistry.java
tsegismont: Ok then

src/main/java/io/vertx/micrometer/impl/VertxHttpClientMetrics.java
jotak: What do you mean? The implementation here differs slightly from the dropwizard one (iirc I used more the hawkular impl as a reference), but endpoints are stored as labels in request count (see line 88 on `requestBegin`)

src/main/java/io/vertx/micrometer/impl/Labels.java
jotak: Not necessarily, the enclosing is final because it's an utility class with only static methods, but this one isn't.

src/main/java/io/vertx/micrometer/impl/VertxMetricsImpl.java
jotak: In a previous discussion with @vietj we were wondering if timer metrics were actually needed and they may even be deprecated / removed from the interface. So, I don't know. An issue to implement it, or an issue in vertx-core to deprecate it?

src/main/java/io/vertx/micrometer/impl/VertxMetricsImpl.java
jotak: I think it's an open debate and I'm from the other side :)

As long as the object is not going to be serialized or such I don't really see why we couldn't use what is more convenient (and safer when used in a side-effect free way ie. non imperative).
Ok it adds a level of indirection, but its usage here is not as such critical for performances as to worry about it, right?

But,  - sigh - I don't necessarily want to enter this debate now :) so if you want me to remove them i'll remove them

src/main/java/io/vertx/micrometer/MicrometerMetricsOptions.java
jotak: done

src/main/java/io/vertx/micrometer/impl/VertxMetricsImpl.java
tsegismont: Ha. Missed that. Yes please go ahead and file an issue on `eclipse/vert.x`

src/main/java/io/vertx/micrometer/impl/VertxMetricsImpl.java
tsegismont: >I think it's an open debate and I'm from the other side :)

The debate is on whether we should start to use `Optional` as return values where it makes sense. For internal code (and fields in particular) there is no debate ;-)

src/main/java/io/vertx/micrometer/impl/Labels.java
tsegismont: :+1:

src/main/java/io/vertx/micrometer/impl/AbstractMetrics.java
tsegismont: Yeah, makes sense :+1:

src/main/java/io/vertx/micrometer/impl/VertxMetricsImpl.java
jotak: Issue created: https://github.com/eclipse/vert.x/issues/2352

src/main/java/io/vertx/micrometer/impl/VertxHttpClientMetrics.java
tsegismont: ok fine

tsegismont:
We're all good! Great job @jotak !

jotak:
Champagne pour moi :)

Issue #25, opened 2018-04-17
Gauges reporting NaN randomly in tests


jotak:
Occasionally we can see test failures due to some gauges reporting NaN instead of the expected double value.
At the moment I am unable to reproduce, but I leave here some information found while debugging:

- Seen in test `VertxNetClientServerMetricsTest.shouldReportNetClientMetrics` : `vertx.net.client.connections` value is NaN
- This gauge is backed by a `LongAdder`, itself stored in a Map in `Gauges`. It seems that from that place, the value is valid at the time of error, and is `0` as expected.
- However when looking at memory directly from micrometer registry, the gauge lambda returns NaN
- Deeper investigation suggests that the `WeakReference` stored in class `DefaultGauge` (micrometer) holds `null` (which is weird since the reference is still in `Gauges` map). In that case `DefaultGauge` returns `0.0D / 0.0` (NaN)

PR #26, opened 2018-04-17
Fix several test issues, potential race cond in Gauges.java, …

jotak:
…stopping registry push jobs

- It may fix issue #25 (bug not reproduced until now)
- better cleaning in tests (release registries after use), and fixing several side-effect exceptions in tests that were not affecting results (such as never-canceled timer)
- Influx and JMX registries now correctly stopped

### DIFF
--- a/src/test/java/io/vertx/micrometer/backend/CustomMicrometerMetricsITest.java
+++ b/src/test/java/io/vertx/micrometer/backend/CustomMicrometerMetricsITest.java
@@ -28,7 +28,6 @@ import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.VertxUnitRunner;
 import io.vertx.micrometer.MetricsDomain;
 import io.vertx.micrometer.MicrometerMetricsOptions;
-import io.vertx.micrometer.backends.BackendRegistries;
 import org.junit.After;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -44,11 +43,12 @@ import static org.assertj.core.api.Assertions.assertThat;
 public class CustomMicrometerMetricsITest {
 
   private static final String REGITRY_NAME = "CustomMicrometerMetricsITest";
+  private Vertx vertx;
   private Vertx vertxForSimulatedServer = Vertx.vertx();
 
   @After
   public void after(TestContext context) {
-    BackendRegistries.stop(REGITRY_NAME);
+    vertx.close(context.asyncAssertSuccess());
     vertxForSimulatedServer.close(context.asyncAssertSuccess());
   }
 
@@ -86,7 +86,7 @@ public class CustomMicrometerMetricsITest {
       }
     }, Clock.SYSTEM));
 
-    Vertx vertx = Vertx.vertx(new VertxOptions()
+    vertx = Vertx.vertx(new VertxOptions()
       .setMetricsOptions(new MicrometerMetricsOptions()
         .setMicrometerRegistry(myRegistry)
         .setRegistryName(REGITRY_NAME)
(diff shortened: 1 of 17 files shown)


src/test/java/io/vertx/micrometer/backend/JmxMetricsITest.java
```diff
@@ -37,18 +37,20 @@
...
   @After
   public void tearDown() {
-    BackendRegistries.stop(MicrometerMetricsOptions.DEFAULT_REGISTRY_NAME);
+    BackendRegistries.stop(REGISTRY_NAME);
   }
 
   @Test
   public void shouldReportJmx(TestContext context) throws Exception {
-    vertx = Vertx.vertx(new VertxOptions()
+    Vertx vertx = Vertx.vertx(new VertxOptions()
```
tsegismont: If you want to make sure everything is closed fine and cannot interfere with other tests, store the `vertx` instance as a field and close it in `tearDown`. You can use `asyncAssertSuccess` with `TestContext` to make sure new tests are not started before Vert.x is closed properly.

I know this is not 100% useful here since there's just one test, but it's a good pattern IMO

src/test/java/io/vertx/micrometer/backend/JmxMetricsITest.java
```diff
@@ -37,18 +37,20 @@
 @RunWith(VertxUnitRunner.class)
 public class JmxMetricsITest {
 
-  private Vertx vertx;
+  private static final String REGISTRY_NAME = "jmx-test";
```
tsegismont: You could use the junit `TestName` rule to make sure each test use a different registry. Again, not 100% useful here, but a good pattern IMO.

src/test/java/io/vertx/micrometer/backend/CustomMicrometerMetricsITest.java
```diff
@@ -69,32 +86,13 @@ public boolean autoCreateDb() {
...
         .setMicrometerRegistry(myRegistry)
         .setRegistryName(REGITRY_NAME)
         .addDisabledMetricsCategory(MetricsDomain.HTTP_SERVER)
         .addDisabledMetricsCategory(MetricsDomain.NAMED_POOLS)
         .setEnabled(true)));
-  }
-
-  @After
-  public void after(TestContext context) {
-    vertx.close(context.asyncAssertSuccess());
```
tsegismont: why not waiting for close any more?

jotak: it's still closing, just the github diff is a little bit screwed up because a bloc of code has been moved from @ Before to the beginning of the test. But I'll change it back to @ Before and make vertx a class field, as you suggested.

src/test/java/io/vertx/micrometer/backend/JmxMetricsITest.java
jotak: done

src/test/java/io/vertx/micrometer/backend/JmxMetricsITest.java
jotak: If you don't mind I prefer to keep it as straightforward as possible until it becomes necessary to add any rule.

tsegismont: No problem

src/test/java/io/vertx/micrometer/backend/JmxMetricsITest.java
tsegismont: :+1:

tsegismont:
@jotak good job!

PR #33, opened 2018-06-05
This annotation is referred from the public API so it requires the ve…

pmlopes:
…rtxgen annotation


jotak:
I'm not knowledgeable enough about this vertx annotations, @vietj can you review?

jotak:
Thanks @pmlopes

Issue #36, opened 2018-06-20
Vertx - InfluxDB metrics keeping connection open


CountZukula:
I am trying to incorporate metrics reporting into my Vertx application (Java), through an InfluxDB instance. However, when I try to gracefully end my application, there is a hanging thread somewhere. I managed to track it down to the InfluxDB connection for vertx's metrics reporting, or so it seems, but I don't see a way to kill it off. Can anyone point me in the right direction?

A minimal working example (if you disable the metrics, the application gracefully finishes, otherwise, it hangs):

````
import io.vertx.core.Vertx;
import io.vertx.core.VertxOptions;
import io.vertx.micrometer.MicrometerMetricsOptions;
import io.vertx.micrometer.VertxInfluxDbOptions;
import io.vertx.micrometer.backends.BackendRegistries;

public class MWE {
    public static void main(String[] args) {
        //setting up the metric options for influxdb. seems to work in MWE without credentials
        MicrometerMetricsOptions metricsOptions = new MicrometerMetricsOptions()
                .setRegistryName("data-client")
                .setInfluxDbOptions(new VertxInfluxDbOptions()
                        //disabling this would make sure the application _does_ gracefully exit
                        .setEnabled(true)
                )
                .setEnabled(true);

        //setting up the vertx instance
        Vertx vertx = Vertx.vertx(
                new VertxOptions()
                        .setMetricsOptions(metricsOptions)
                        .setWorkerPoolSize(50)
        );

        //stop vertx after a second
        vertx.setTimer(1000, timerID -> {
            //closing the vertx instance
            vertx.close(result -> System.out.println("Vertx was closed."));
            //closing the registry of metrics to influxdb
            BackendRegistries.getNow("data-client").close();
            System.out.println("Closed everything");
        });

        System.out.println("Done with main");
    }
}
````

(originally posted here: https://stackoverflow.com/questions/50945351/vertx-influxdb-metrics-keeping-connection-open )

Some relevant dependencies:

````
        <dependency>
            <groupId>io.vertx</groupId>
            <artifactId>vertx-core</artifactId>
            <version>3.5.2</version>
        </dependency>

        <dependency>
            <groupId>io.vertx</groupId>
            <artifactId>vertx-web</artifactId>
            <version>3.5.2</version>
        </dependency>

        <dependency>
            <groupId>io.vertx</groupId>
            <artifactId>vertx-rx-java2</artifactId>
            <version>3.5.2</version>
        </dependency>

[...]

PR #37, opened 2018-06-20
Update micrometer to 1.0.5, fix broken tests/code and update doc

jotak:
In Influx registry, call `close` method instead of `stop` as it will release the executor.


jotak:
Fixes #36

src/main/asciidoc/index.adoc
```diff
@@ -323,6 +323,11 @@ Finally it is possible to filter the returned metrics from their base names:
 
 This section lists all the metrics generated by monitoring the Vert.x core tools.
 
+A word about naming: the different metric backends may have different conventions or rules for naming metrics.
```
tsegismont: Replace `A word about naming:` with `NOTE:`

tsegismont: Replace `A word about naming:` with `NOTE:`

jotak: Done

Issue #39, opened 2018-06-28
Filter out internal eventbus addresses


jotak:
Once https://github.com/eclipse/vert.x/pull/2516 is merged, internal eventbus addresses will be distinguished with pattern "__vertx.[net|ws|reply].randomUUID".

More generally, we can assume that addresses starting with "__vertx" are internal.
They should be filtered out from eventbus metrics.

Open question: any interest in making this filtering configurable? ie. user will be able to track vertx internal eventbus messages in metrics?

tsegismont:
@jotak I think it's fine to simply ignore them for now (no filtering flag)

PR #40, opened 2018-07-09
Ignore internal eventbus metrics

jotak:
See issue #39

### DIFF
--- a/src/main/java/io/vertx/micrometer/impl/VertxEventBusMetrics.java
+++ b/src/main/java/io/vertx/micrometer/impl/VertxEventBusMetrics.java
@@ -58,15 +58,25 @@ class VertxEventBusMetrics extends AbstractMetrics implements EventBusMetrics<Ve
     bytesWritten = summaries("bytesWritten", "Number of bytes sent while sending messages to event bus cluster peers", Label.EB_ADDRESS);
   }
 
+  private static boolean isInternal(String address) {
+    return address.startsWith("__vertx.");
+  }
+
   @Override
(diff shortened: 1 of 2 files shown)


jotak:
@vietj / @tsegismont if it's ok with you I'll do a similar patch in dropwizard module

Ps: do you think there's any need in adding a config flag to turn on/off ignoring internal EB addresses?

vietj:
@jotak that would be great

src/main/java/io/vertx/micrometer/impl/VertxEventBusMetrics.java
```diff
@@ -58,15 +58,25 @@
...
   }
 
+  private static boolean isInternal(String address) {
+    return address.startsWith("__vertx.");
+  }
+
   @Override
   public Handler handlerRegistered(String address, String repliedAddress) {
+    if (isInternal(address)) {
+      // Ignore interna metrics
```
tsegismont: Typo _interna**l**_

src/main/java/io/vertx/micrometer/impl/VertxEventBusMetrics.java
```diff
@@ -137,5 +161,13 @@ public void close() {
     Handler(String address) {
       this.address = address;
     }
+
+    static Handler ignored() {
+      return new Handler(null);
```
tsegismont: How about returning a constant instead of creating a new instance?

jotak:
@tsegismont done

tsegismont:
Thanks @jotak !

Issue #41, opened 2018-07-24
Provide HTTP Method key in all the relevant VertxHttpServerMetrics


mangatmodi:
For following Metrics, we don't have HTTP Method Key.
  - requests
  - requestResetCount
  - responseTime

It is not possible to have correct analysis without the Method. For e.g. response time for GET and POST on a resource might differ. We are already having HTTP Method available in requestCount, so same can be easily done for others.

mangatmodi:
I have created a PR for the same here https://github.com/vert-x3/vertx-micrometer-metrics/pull/42.

tsegismont:
@jotak make sure to add the milestone on issues fixed in master

PR #42, opened 2018-07-24
[#41] Provide HTTP Method key in all the relevant VertxHttpServerMetrics 

mangatmodi:
Add Label.Method for following Vertx HTTP server metrics. The PR fixes https://github.com/vert-x3/vertx-micrometer-metrics/issues/41. 
 - requests
 - requestResetCount
 - responseTime


jotak:
Thanks @mangatmodi , good change indeed!

For consistency, it would be nice to have the same on client metrics (class `VertxHttpClientMetrics`). Do you mind making the same changes there as well? (it's almost identical)

mangatmodi:
Sure I will do the same :)

PR #43, opened 2018-08-16
Increment responseCount counter

kerko:
The counters for "responseCount" are retrieved but never incremented.
I dont know if this should also be backported to the 3.5 branch.


vietj:
it would be great to have a test for this

jotak:
Hi @kerko ,
Thanks for this patch! For your information, there's already a test for client metrics, which just doesn't correctly cover the response counter: see this line:
https://github.com/vert-x3/vertx-micrometer-metrics/blob/master/src/test/java/io/vertx/micrometer/VertxHttpClientServerMetricsTest.java#L109
that could be removed and replaced with a new check in the above assertion like:

`dp("vertx.http.client.responseCount[code=200,local=?,method=POST,path=/resource,remote=127.0.0.1:9195]$COUNT", concurrentClients * HTTP_SENT_COUNT)`

kerko:
I added a test a verified it works as intended

Issue #52, opened 2018-10-30
Quantiles not available in metrics


cescoffier:
I've a simple application explosing a HTTP endpoint (greeting endpoint):

```java
public class MyVertxApp extends AbstractVerticle {

    public static void main(String[] args) {
        Vertx vertx = Vertx.vertx(new VertxOptions().setMetricsOptions(
                new MicrometerMetricsOptions()
                        .setPrometheusOptions(new VertxPrometheusOptions().setEnabled(true)
                                .setStartEmbeddedServer(true)
                                .setEmbeddedServerOptions(new HttpServerOptions().setPort(8081))
                                .setEmbeddedServerEndpoint("/metrics"))
                        .setEnabled(true)));

        vertx.deployVerticle(MyVertxApp.class.getName());
    }

    @Override
    public void start() {

        vertx.createHttpServer()
                .requestHandler(req -> req.response().end("Hello"))
                .listen(8080);
    }
```

The promethous output is quite minimal:

```
vertx_http_server_requests{local="0.0.0.0:8080",path="/",remote="_",} 0.0
vertx_http_server_requests{local="0.0.0.0:8081",path="/metrics",remote="_",} 1.0
# HELP vertx_verticle_deployed Number of verticle instances deployed
# TYPE vertx_verticle_deployed gauge
vertx_verticle_deployed{class="org.acme.MyVertxApp",} 1.0
# HELP vertx_http_server_connections Number of opened connections to the server
# TYPE vertx_http_server_connections gauge
vertx_http_server_connections{local="0.0.0.0:8080",remote="_",} 0.0
vertx_http_server_connections{local="0.0.0.0:8081",remote="_",} 2.0
# HELP vertx_http_server_requestCount_total Number of processed requests
# TYPE vertx_http_server_requestCount_total counter
vertx_http_server_requestCount_total{code="200",local="0.0.0.0:8080",method="GET",path="/",remote="_",} 52.0
vertx_http_server_requestCount_total{code="200",local="0.0.0.0:8081",method="GET",path="/metrics",remote="_",} 30.0
# HELP vertx_http_server_bytesSent Number of bytes sent by the server
# TYPE vertx_http_server_bytesSent summary
vertx_http_server_bytesSent_count{local="0.0.0.0:8080",remote="_",} 52.0
vertx_http_server_bytesSent_sum{local="0.0.0.0:8080",remote="_",} 260.0
[...]

PR #53, opened 2018-11-05
Add documentation for percentiles

jotak:
cc @cescoffier => related to https://github.com/vert-x3/vertx-micrometer-metrics/issues/52

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -197,6 +197,40 @@ In this example, metrics are reported both for JMX and Graphite:
 
 Please refer to {@link io.vertx.micrometer.MicrometerMetricsOptions} for an exhaustive list of options.
 
+=== Averages and quantiles in Prometheus
+
+By default, when using the Prometheus registry, histogram-kind metrics will not contain averages or quantile stats.
+
+Averages don't come out of the box but they are typically link:https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations[computed at query time],
+with `promql`. Example, for HTTP client response time average during the last 5 minutes:
+
+[source]
+----
+  rate(vertx_http_client_responseTime_seconds_sum[5m])
+/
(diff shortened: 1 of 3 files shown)


cescoffier:
Great! 

Wondering if we can avoid the downcasting of the registry, something like: `withPercentiles` that would to the work.

jotak:
You mean, using our own Metrics options instead above the micrometer's registry?

Currently our options are just wrapping the micrometer registry and provide some handful shortcuts for some micrometer internal configs, but it's not 1-to-1 (many micrometer options are not reflected into vertx metrics options) so in many cases it's still necessary to have a direct access to the micrometer registry.

The advantage of configuring directly the micrometer registry is that you'll have more options / more control. Ex: here in the example I show how to activate all percentiles, which leads to many additional timeseries that are stored, but a user could fine-tune with only a smaller subset, like limited to 0.95 and 0.99, which will save a lot of space:

```
    registry.config().meterFilter(
        new MeterFilter() {
          @Override
          public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticConfig config) {
            return DistributionStatisticConfig.builder()
                .percentiles(0.95, 0.99)
                .build()
                .merge(config);
          }
        });
```

And there's still other options like slas... so at some point I prefer to guide the user through fine-tuning the micrometer registry (and delegate the responsibility of configuration to it) so that he gets full control of it.

jotak:
As it's only doc & example update, is it ok to merge now ? (regarding 3.6 release)
Needs a 3.6 branch or is it fine to put only in master? (cc @vietj )

tsegismont:
@jotak I'm fine with merging. But I believe it would be useful to have a switch (disabled by default) in options to enable statistics with sensible defaults. This documentation will not be wasted as it will show how to proceed for fine tuning.

jotak:
@tsegismont so i'm merging, and working on a separate PR for the option

Issue #59, opened 2018-11-09
ExternalConfigurationTest.testPrometheusDefaults failure in CI


vietj:
reported by Travis : https://travis-ci.org/vert-x3/vertx-micrometer-metrics/jobs/452894113

```
testPrometheusDefaults(io.vertx.micrometer.ExternalConfigurationTest)  Time elapsed: 0.459 sec  <<< ERROR!
io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:9999
	at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
	at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:717)
	at io.netty.channel.socket.nio.NioSocketChannel.doFinishConnect(NioSocketChannel.java:325)
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:340)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:634)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:581)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:498)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:460)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:884)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.net.ConnectException: Connection refused
	at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
	at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:717)
	at io.netty.channel.socket.nio.NioSocketChannel.doFinishConnect(NioSocketChannel.java:325)
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:340)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:634)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:581)
[...]

PR #64, opened 2018-11-22
Tests: retry mechanism while trying to connect to running prometheus …

jotak:
…server

Hopefully should fix issue #59

- Create class PrometheusTestHelper to mutualize retry mechanism across tests
- Use assertj in ExternalConfigurationTest (provides better test logs on failures - faster debugging)

### DIFF
--- a/src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
+++ b/src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
@@ -33,20 +33,45 @@ import java.util.Set;
  * @author Joel Takvorian
  */
 public final class PrometheusTestHelper {
+  private static final int DEFAULT_MAX_ATTEMPS = 10;
+  private static final long DEFAULT_SLEEP_BEFORE_RETRY_MS = 100;
+
   private PrometheusTestHelper() {
   }
 
   public static void tryConnect(Vertx vertx, TestContext context, int port, String host, String requestURI, Handler<Buffer> bodyReader) {
-    tryConnect(vertx, port, host, requestURI, res -> {
+    tryConnect(vertx, context, port, host, requestURI, bodyReader, DEFAULT_MAX_ATTEMPS, DEFAULT_SLEEP_BEFORE_RETRY_MS);
+  }
+
+  public static void tryConnect(Vertx vertx,
+                                TestContext context,
+                                int port,
+                                String host,
+                                String requestURI,
+                                Handler<Buffer> bodyReader,
+                                int maxAttempts,
+                                long sleepBeforeRetryMs) {
+    tryConnect(vertx, context, port, host, requestURI, res -> {
       context.assertEquals(200, res.statusCode());
       res.bodyHandler(bodyReader);
-    }, 0);
+    }, maxAttempts, sleepBeforeRetryMs, 0);
   }
 
   public static Set<String> getMetricNames(Vertx vertx, TestContext context, int port, String host, String requestURI, long timeout) {
+    return getMetricNames(vertx, context, port, host, requestURI, timeout, DEFAULT_MAX_ATTEMPS, DEFAULT_SLEEP_BEFORE_RETRY_MS);
+  }
+
+  public static Set<String> getMetricNames(Vertx vertx,
+                                           TestContext context,
+                                           int port,
+                                           String host,
+                                           String requestURI,
+                                           long timeout,
+                                           int maxAttempts,
+                                           long sleepBeforeRetryMs) {
     Async async = context.async();
     Set<String> metrics = Collections.synchronizedSet(new HashSet<>());
-    tryConnect(vertx, port, host, requestURI, resp -> {
+    tryConnect(vertx, context, port, host, requestURI, resp -> {
       context.assertEquals(200, resp.statusCode());
       RecordParser parser = RecordParser.newDelimited("\n", resp);
       parser.exceptionHandler(context::fail).endHandler(v -> {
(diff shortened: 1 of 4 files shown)


tsegismont:
@jotak can you update `testJvmMetricsEnabled` ? It seems some JVM metric names changed.

jotak:
My bad ... I was doing different things at the same time and didn't split commits correctly.

jotak:
@tsegismont can you look at it again please?

src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
```diff
@@ -0,0 +1,90 @@
...
+                                 String host,
+                                 String requestURI,
+                                 Handler<HttpClientResponse> respHandler,
+                                 int attempt) {
+    HttpClientRequest req = vertx.createHttpClient()
+      .get(port, host, requestURI)
+      .handler(respHandler)
+      .exceptionHandler(e -> {
+        if (attempt < 10) {
+          System.out.println(e);
```
tsegismont: Please use debug logging instead of stdout. Our CI is sensitive to long outputs (build failed at some point). With logging output becomes configurable.

src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
```diff
@@ -0,0 +1,90 @@
...
+          try {
+            Thread.sleep(500);
+          } catch (InterruptedException e1) {
+            e1.printStackTrace();
+          }
+          System.out.println("retrying...");
```
tsegismont: Same as above, please remove or log debug

src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
```diff
@@ -0,0 +1,90 @@
...
+          tryConnect(vertx, port, host, requestURI, respHandler, attempt + 1);
+        } else {
+          System.out.println("aborting");
```
tsegismont: Same here

src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
```diff
@@ -0,0 +1,90 @@
...
+                                 Handler<HttpClientResponse> respHandler,
+                                 int attempt) {
+    HttpClientRequest req = vertx.createHttpClient()
+      .get(port, host, requestURI)
+      .handler(respHandler)
+      .exceptionHandler(e -> {
+        if (attempt < 10) {
+          System.out.println(e);
+          try {
+            Thread.sleep(500);
```
tsegismont: Do not block the event loop :) `vertx.setTimer` would do just fine here.

Besides it would be better to try more often (e.g every 100ms) and have a configurable timeout.

src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
```diff
@@ -0,0 +1,90 @@
...
+    }, 0);
+    async.await(timeout);
+    return metrics;
+  }
+
+  private static void tryConnect(Vertx vertx,
+                                 int port,
+                                 String host,
+                                 String requestURI,
+                                 Handler<HttpClientResponse> respHandler,
```
tsegismont: If you make this a `Handler<AsyncResult>` then you'll be able to manage failures properly (iiuc the current code just prints `aborting`)

jotak: Hmm not sure what you mean, `respHandler` is actually used by caller to read the HTTP response.

src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
jotak: Oh yes, vert.x mantra :)
(I hoped it was ok for a small little test :) )

src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
tsegismont: This handler is currently only invoked in case of success. Otherwise, the test prints "aborting" and then times out.

If instead of a simple handler you have a handler of async result, then you can fail the test properly with a reason message.

jotak:
@tsegismont should be better now

src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
```diff
@@ -67,22 +95,20 @@ private static void tryConnect(Vertx vertx,
...
-          System.out.println(e);
-          try {
-            Thread.sleep(500);
-          } catch (InterruptedException e1) {
-            e1.printStackTrace();
-          }
-          System.out.println("retrying...");
-          tryConnect(vertx, port, host, requestURI, respHandler, attempt + 1);
+        if (attempt < maxAttempts) {
+          LOGGER.error("Connection failed (retrying...)", e);
```
tsegismont: if was thinking about a simple debug log given it's expected that it won't succeed right away

src/test/java/io/vertx/micrometer/backend/PrometheusTestHelper.java
jotak: ok I think I understand

jotak: @tsegismont I've updated with a different solution, tell me what you think. Just passing the testcontext into `tryConnect` method, and failing properly, and removed logs.

TBH I've twisted my mind to try with AsyncResult, but I might not be familiar enough with it (I imagined it was some kind of future, but it seems i was wrong). If you think it's still better to use it then I'd appreciate some help :)

tsegismont:
@jotak the build is still failing

jotak:
@tsegismont now green

tsegismont:
@jotak thanks, will review latest changes this morning

tsegismont:
@jotak good job, thanks!

Issue #71, opened 2019-03-20
Using custom Prometheus collector registry


nolequen:
In my application I have own Prometheus metrics and I want to use the same collector registry as for Vert.x metrics.

jotak:
Hi @nolequen 
I think you can already do that with `MicrometerMetricsOptions.setMicrometerRegistry`.
Looking at your commit, what you want to do is to re-use not the _micrometer_ registry, but the _collector registry_ from Prometheus client (without micrometer).

I guess you could do something like this, while setting up your vert.x options:

```java
    PrometheusMeterRegistry micrometerRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT, myCollectorRegistry, Clock.SYSTEM);
    metricsOptions.setMicrometerRegistry(micrometerRegistry);
```

without any change in vertx code. Does it sound good?

nolequen:
hi @jotak 
Your solution looks nice, but unfortunately it ignores specified VertxPrometheusOptions.

jotak:
@nolequen that's right ... so you would have to basically duplicate this code within your application code: https://github.com/vert-x3/vertx-micrometer-metrics/blob/master/src/main/java/io/vertx/micrometer/backends/PrometheusBackendRegistry.java

I'm going to patch so don't you don't have to.

nolequen:
@jotak why you want to patch by yourself? why not to merge the PR?

jotak:
@nolequen since there's already a mechanism to provide a registry, I'm not sure that adding a second one is really necessary. It just has to be fixed regarding the issue you raised (and this, regardless if your PR is merged or not). In terms of API, having different ways to achieve the same goal may be confusing.

nolequen:
@jotak ok. I'll be waiting for the fix. thanks!

PR #73, opened 2019-03-21
Fix reused registry ignoring Prometheus options

jotak:
Also add test, example & documentation

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -281,7 +281,7 @@ link:http://micrometer.io/docs/concepts#_registry[Micrometer doc].
 
 === Reusing an existing registry
 
-It is possible to reuse an existing Micrometer registry (or `CollectorRegistry` from the Prometheus java client),
+It is possible to reuse an existing Micrometer registry (or `CollectorRegistry` from the Prometheus Java client),
 and inject it into the Vert.x metrics options:
 
 [source,$lang]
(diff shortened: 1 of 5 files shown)


jotak:
Fixes #71 

@nolequen thank you for your PR, but as said, I think it's better to not add multiple different ways to inject an existing registry, unless there's good reasons for that. Does that sound good to you?

src/main/asciidoc/index.adoc
```diff
@@ -279,7 +279,17 @@ As an example, here is a custom timer that will track the execution time of a pi
 For more examples, documentation about the Micrometer registry and how to create metrics, check
 link:http://micrometer.io/docs/concepts#_registry[Micrometer doc].
 
-=== Other instrumentation
+=== Reusing an existing registry
+
+It is possible to reuse an existing Micrometer registry (or `CollectorRegistry` from the Prometheus java client),
```
tsegismont:
```suggestion
It is possible to reuse an existing Micrometer registry (or `CollectorRegistry` from the Prometheus Java client),
```

src/main/java/examples/MicrometerMetricsExamples.java
```diff
@@ -246,4 +248,22 @@ public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticC
           }
         });
   }
+
+  public void useExistingRegistry() {
+    // This registry might be used to collect metrics other than Vert.x ones
+    PrometheusMeterRegistry registry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
+
+    // You could also reuse an existing registry from the Prometheus java client:
```
tsegismont:
```suggestion
    // You could also reuse an existing registry from the Prometheus Java client:
```

tsegismont:
@jotak after it's merged, can you please cherry-pick the commit to `3.7` branch? Thank you

PR #77, opened 2019-04-02
fix(NPE): Add null check for tag values

jahanzebbaber:
Naming is hard :(
Sorry for messing up your git history.


jotak:
no problem, namings are not always perfect english especially while naming test methods :)

Issue #79, opened 2019-05-17
java.lang.NoClassDefFoundError: io/micrometer/prometheus/PrometheusMeterRegistry


hc-codersatlas:
Hi, getting java.lang.NoClassDefFoundError: io/micrometer/prometheus/PrometheusMeterRegistry

Because it's required in BackendRegistries.

I'm not using Prometheus  and so haven't imported it in micrometer. Specifying a custom meterRegistry as per documentation.

manzur:
Hey, 

I had a similar issue, so made a band aid fix. As mentioned in the commit message, branching can be re-written in a different way, as the option with specifying both registry and additional options seems to be not mentioned in the docs.

jotak:
Thank you @hc-codersatlas , @manzur for spotting the issue. @manzur I'm reviewing your PR.

PR #80, opened 2019-05-18
reverse condiditional check for configuring Prometheus backend

manzur:
Having a check of `instanceof` before `#getPrometheusOptions` causes
code using other backend registry throw no class defined exception:

```
java.lang.NoClassDefFoundError: io/micrometer/prometheus/PrometheusMeterRegistry
```

Reversing this condition fixes the issue due to both short-circuiting and lazy class loading.

Generally `instanceof` branch can be removed at all as 2nd PrometheusBackendRegistry constructor
does nothing else than setting generic prometheus options that could be done in:

https://github.com/vert-x3/vertx-micrometer-metrics/blob/74d70d238d1c160318bbf5f558c39e64af74b679/src/main/java/io/vertx/micrometer/backends/PrometheusBackendRegistry.java#L51

Original issue reference: https://github.com/vert-x3/vertx-micrometer-metrics/issues/79


jotak:
I'll cherry-pick to 3.7 branch

Issue #92, opened 2020-01-07
Label matches don't seem to actually filter anything.


sfitts:
Adding a label match would seem to imply that if a given meter has a non-matching value for the given label, it should be rejected.  However, the current implementation is using these to create Micrometer "accept" filters.  All those can do is guarantee acceptance.  This means that for non-matching meters the response is "NEUTRAL" which by default means the meter will be added.

At a minimum it seems that there needs to be a way to specify that a given label match should be used to create a "deny" filter instead (thus allowing for a non-match to be "denied").

jotak:
Oh, that's right! We need to reverse the logic
https://micrometer.io/docs/concepts#_denyaccept_meters

jotak:
For anybody who'd like a workaround before the fix is merged, it's possible to achieve the same directly via micrometer registry config:

```java
    MeterRegistry registry = BackendRegistries.getDefaultNow();
    registry.config().meterFilter(
      MeterFilter.deny(id -> {
        String tagValue = id.getTag("label-name");
        return !"label-value".equals(tagValue);
      }));
```

PR #93, opened 2020-01-13
Fix acceptance matchers ineffective

jotak:
Need to reverse logic and use "deny" instead of "accept" ; "accept" only is useful to shortcut next filters processing but will not be useful to filter out metrics (unless they're explicitly said to be all denied by default)

Added a test

Fixes https://github.com/vert-x3/vertx-micrometer-metrics/issues/92

### DIFF
--- a/src/main/java/io/vertx/micrometer/backends/BackendRegistries.java
+++ b/src/main/java/io/vertx/micrometer/backends/BackendRegistries.java
@@ -125,13 +125,13 @@ public final class BackendRegistries {
         case EQUALS:
           if (m.getAlias() == null) {
             // Exact match => accept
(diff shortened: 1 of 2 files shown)


jotak:
cc @sfitts

tsegismont:
@jotak can you check the build failure?

jotak:
@tsegismont yes, see https://github.com/vert-x3/vertx-micrometer-metrics/pull/94 ; maybe there's a bug in vertx-core actually?

jotak:
@tsegismont build is passing now

jotak:
@tsegismont cherry-picked on 3.8

Issue #95, opened 2020-02-11
"Pending" message gauges should be set to zero when handler is unregistered


sfitts:
### Version

3.8.4

### Context

When looking at the values reported for the `vertx_eventbus_pending` gauge associated with the "credit" EB addresses I noticed that they frequently got "stuck" with a non-zero value.  These addresses are generated and tend to come and go quickly, so it isn't uncommon for one to be unregistered while it still has a pending message to process (at least not at load).

Currently the Metrics SPI does not report messages that are "discarded" during the unregister process (see https://github.com/eclipse-vertx/vert.x/issues/3281), so until it does the metrics provider will either need to remove these gauges entirely or at least reset them to 0 in the `handlerUnregistered` routine.

### Steps to reproduce

Simply running an application which uses a message producer (and thus the credit side-channel) at any reasonable scale will illustrate the issue.  You can also see that in `HandlerRegistration.java` between lines 168 and 177, any pending messages are discarded, but the SPI is not called (mainly because it lacks the entry point).

jotak:
It makes sense to me, either to have a new hook for discarded messages, or to mirror the behaviour of `HandlerRegistration` and reset pending list when it's unregistered.

@vietj or @tsegismont do you agree?

I also see that in Vert.x 4 the `HandlerRegistration` class seems to have been widely refactored so I'm not sure if this needs to be done in master as well? (I don't see any discarding mechanism anymore there)

tsegismont:
@jotak I commented on the PR. I believe it makes sense to reset pending list in the SPI implementation when the `handlerUnregistration` callback is invoked

tsegismont:
Fixed by a385f82

vietj:
see https://github.com/vert-x3/vertx-dropwizard-metrics/issues/97 and https://github.com/vert-x3/vertx-dropwizard-metrics/issues/96

PR #98, opened 2020-02-18
Reset pending gauge on handler unregistered

jotak:
Mirroring behaviour in HandlerRegistration
Fixes https://github.com/vert-x3/vertx-micrometer-metrics/issues/95

### DIFF
--- a/src/test/java/io/vertx/micrometer/VertxEventBusMetricsTest.java
+++ b/src/test/java/io/vertx/micrometer/VertxEventBusMetricsTest.java
@@ -137,7 +137,7 @@ public class VertxEventBusMetricsTest {
         consumerRef.set(consumer);
         step1EBReady.complete();
       }
-    }, new DeploymentOptions());
+    }, new DeploymentOptions().setWorker(true));
     step1EBReady.awaitSuccess();
 
     // Send to eventbus
--- a/src/test/java/io/vertx/micrometer/VertxEventBusMetricsTest.java
+++ b/src/test/java/io/vertx/micrometer/VertxEventBusMetricsTest.java
@@ -1,11 +1,7 @@
 package io.vertx.micrometer;
 
-import io.vertx.core.AbstractVerticle;
-import io.vertx.core.DeploymentOptions;
-import io.vertx.core.Future;
-import io.vertx.core.Promise;
-import io.vertx.core.Vertx;
-import io.vertx.core.VertxOptions;
+import io.vertx.core.*;
+import io.vertx.core.eventbus.MessageConsumer;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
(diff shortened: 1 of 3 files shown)


jotak:
Now I don't know if this needs to be cherry-picked to master ... as it seems in master vertx-core we are not discarding the pending events on unregister? (or maybe it's still done but in a different place, I couldn't find that)

tsegismont:
@jotak would you mind to check if there's a similar issue with the Dropwizard implementation? Thank you

jotak:
@tsegismont 
About dropwizard, I believe there's nothing to do: see https://github.com/vert-x3/vertx-dropwizard-metrics/blob/master/src/main/java/io/vertx/ext/dropwizard/impl/EventBusMetricsImpl.java#L251-L253
Pending gauges are already updated on unregistering.

About adding a test: I struggle to make something reproducible about pending messages. I can easily check that after registration there's no pending message, but that wouldn't prove anything about what we do here. Do you have any suggestion?

tsegismont:
Le mer. 19 févr. 2020 à 10:19, Joel Takvorian <notifications@github.com> a
écrit :

> @tsegismont <https://github.com/tsegismont>
> About dropwizard, I believe there's nothing to do: see
> https://github.com/vert-x3/vertx-dropwizard-metrics/blob/master/src/main/java/io/vertx/ext/dropwizard/impl/EventBusMetricsImpl.java#L251-L253
> Pending gauges are already updated on unregistering.
>

👍


> About adding a test: I struggle to make something reproducible about
> pending messages. I can easily check that after registration there's no
> pending message, but that wouldn't prove anything about what we do here. Do
> you have any suggestion?
>

Create a worker verticle that registers a consumer.
Send a few messages from the test body.
In the consumer body, block the worker thread after the first message is
received
While the worker thread is blocked you can verify there are a few pending
messages.
After the worker thread is unblocked unregister the consumer.

I believe this should do

src/test/java/io/vertx/micrometer/VertxEventBusMetricsTest.java
```diff
@@ -111,4 +106,61 @@ public void start(Promise<Void> future) throws Exception {
...
+
+    // Setup eventbus handler
+    vertx.deployVerticle(() -> new AbstractVerticle() {
+      @Override
+      public void start(Promise<Void> future) {
+        MessageConsumer<Object> consumer = vertx.eventBus().consumer("testSubject", msg -> {
+          if (first.getAndSet(false)) {
+            step2FirstReceived.complete();
+          }
+          step3BlockWorker.await();
```
jotak: @tsegismont I did more or less as you suggested, with a difference: event consuming is blocked on the event loop and not on worker thread. I know it can be a bit risky on a slower environment - here it works on my machine (let's see what CI says).

I tried with a worker thread as well, using:

```
WorkerExecutor executor = vertx.createSharedWorkerExecutor("my-worker-pool", 1);
 // then in consumer handler:
executor.executeBlocking(promise -> { /*block*/... });
```

But it doesn't work, as this is not blocking thequeue processing (pending count is decreased despite blocking, here).

Or is there another way to do, such as telling vertx to use my worker for EB queue processing?

jotak: Actually, I think I understand how to do: now using `.setWorker(true).setWorkerPoolSize(1)` on my deployment options... is that what you expected?

tsegismont: It doesn't work without `.setWorkerPoolSize(1)` ?

jotak: It does ... i thought it would be necessary to have the other events blocked, but it's not. I removed it.

jotak:
thanks @tsegismont , i'll cherry-pick to 3.9 and 4

PR #101, opened 2020-02-24
[4.0] Reset pending gauge on handler unregistered

jotak:
This reports https://github.com/vert-x3/vertx-micrometer-metrics/pull/98 on master / 4.0

@tsegismont some weird thing here: the test written for 3.8 that you reviewed isn't valid here anymore. Something changed in EB behaviour that I don't know if it's intentional or a bug.

You can see in my test, I've added some logs. Here's the output:

```
Await EB ready...
EB ready!
Await first received...
First received!
Await thread unblock...
Await thread unblock...
Await thread unblock...
Await thread unblock...
vertx.eventbus.handlers[]$VALUE/1.0
vertx.eventbus.processed[side=local]$COUNT/4.0
vertx.eventbus.delivered[side=local]$COUNT/4.0
vertx.eventbus.received[side=local]$COUNT/4.0
vertx.eventbus.published[side=local]$COUNT/4.0
vertx.eventbus.pending[side=local]$VALUE/0.0

java.lang.AssertionError: 
Expecting:
 <[vertx.eventbus.handlers[]$VALUE/1.0,
    vertx.eventbus.processed[side=local]$COUNT/4.0,
    vertx.eventbus.delivered[side=local]$COUNT/4.0,
    vertx.eventbus.received[side=local]$COUNT/4.0,
    vertx.eventbus.published[side=local]$COUNT/4.0,
    vertx.eventbus.pending[side=local]$VALUE/0.0]>
to contain:
 <[vertx.eventbus.pending[side=local]$VALUE/3.0]>
but could not find:
 <[vertx.eventbus.pending[side=local]$VALUE/3.0]>
```

Notice the four `Await thread unblock...`: it means blocking a EB consumer thread doesn't block execution of next consumer calls anymore. So all events are consumed even if they're blocked.
Is this a bug or not?

The test would succeed without using a Worker verticle, this is specific to workers. So a quick fix could be to remove use of worker verticle, with a risk to block the event loop.

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -50,29 +50,11 @@ Sets metrics types that are disabled.
 |[[enabled]]`@enabled`|`Boolean`|+++
 Set whether metrics will be enabled on the Vert.x instance. Metrics are not enabled by default.
 +++
-|[[influxDbOptions]]`@influxDbOptions`|`link:dataobjects.html#VertxInfluxDbOptions[VertxInfluxDbOptions]`|+++
-Set InfluxDB options.
- Setting a registry backend option is mandatory in order to effectively report metrics.
-+++
-|[[jmxMetricsOptions]]`@jmxMetricsOptions`|`link:dataobjects.html#VertxJmxMetricsOptions[VertxJmxMetricsOptions]`|+++
-Set JMX metrics options.
- Setting a registry backend option is mandatory in order to effectively report metrics.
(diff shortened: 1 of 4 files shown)


tsegismont:
@vietj I believe you did some changes with respect to handler holders and metrics for 4.0. Does that ring a bell?

jotak:
Closing: will be replaced with https://github.com/vert-x3/vertx-micrometer-metrics/pull/103

Issue #106, opened 2020-06-25
Client metrics implementation


vietj:
This https://github.com/eclipse-vertx/vert.x/issues/3467 provides a client metrics SPI (similar to HTTP client metrics) that will be used by Vert.x clients to report metrics, e.g the SQL client uses it since https://github.com/eclipse-vertx/vertx-sql-client/issues/698

vietj:
@jotak FYI , this should not be too much work

jotak:
What's the expectations for these metrics? I guess counter + processing time measurement for both requests and responses?
Is there an implementation reference (dropwizard...) or not yet?

jotak:
Put differently, what is worth measuring ? Given the example provided, I would say:

- Queue time
- Full processing time w/ network latency (from RQ begin to RS end)

Does RQ processing time alone, or RS time alone make sense to measure? (btw I'm seeing in https://github.com/eclipse-vertx/vertx-sql-client/commit/e8b9e0437d0ae30978d03d203d3f0d90feb45600 that RQ begin is called without RQ end)

vietj:
is there a bug in SQL client implementation ?

vietj:
@jotak see https://github.com/eclipse-vertx/vertx-sql-client/commit/0ff45b2f21ed61c657a1ab20d3c4ff9a04fb7b29 thanks :-)

PR #107, opened 2020-06-26
Implement ClientMetrics interface

jotak:
Creates 5 new metrics per type of client:
- vertx.xxx.queue.time (histogram)
- vertx.xxx.queue.pending (gauge)
- vertx.xxx.processing.time (histogram)
- vertx.xxx.processing.pending (gauge)
- vertx.xxx.reset (counter)

They're all labeled by remote address and namespace. The provided "type" (e.g. "sql") is used as metric prefix (e.g. "vertx.sql.").

There's some refactoring on categories / MetricDomains, as this was before a closed list (enum), now it's an opened list that accepts custom values (e.g. "sql"). Hence it's still possible to turn off just one "ClientMetrics" instance based on its type, via the metrics options.

Fixes https://github.com/vert-x3/vertx-micrometer-metrics/issues/106

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -44,7 +44,7 @@ Set the matched value.
 [frame="topbot"]
 |===
 ^|Name | Type ^| Description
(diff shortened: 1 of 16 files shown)


jotak:
Do not merge: missing tests

PR #110, opened 2020-07-07
Bump micrometer 1.5.2

jotak:
Note that 1.5.x is an LTS release line of Micrometer. It's better to have it for vert.x 4.

Unit tests seem OK. I'll run some end-to-end tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,7 @@
     <assertj.version>3.9.0</assertj.version>
     <junit.version>4.12</junit.version>
     <maven-failsafe-plugin.version>2.18.1</maven-failsafe-plugin.version>
-    <micrometer.version>1.1.0</micrometer.version>
+    <micrometer.version>1.5.2</micrometer.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -59,17 +59,6 @@
       <groupId>io.micrometer</groupId>
       <artifactId>micrometer-core</artifactId>
       <version>${micrometer.version}</version>
-      <exclusions>
```
tsegismont, line 62: Out of curiosity, why did we have this issue with `HdrHistogram`? Dependency convergence?

tsegismont:
Thank you @jotak !

pom.xml
jotak: Some conflicts with another vertx module, but I don't remember which one, perhaps @vietj remembers... but I think it's now safe as micrometers uses a similar minor version now.
If putting all modules together creates again conflicts with HdrHistogram, we'll put this exclusion back.

tsegismont: Fine, thanks for the details.

PR #113, opened 2020-09-02
Fix crash with unix sockets

jotak:
Fixes https://github.com/vert-x3/vertx-micrometer-metrics/issues/112

cc @Stimzz

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,6 +141,14 @@
       <version>1.7.25</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <!-- Test dependency is for Unix socket test -->
+      <groupId>io.netty</groupId>
+      <artifactId>netty-transport-native-epoll</artifactId>
+      <version>4.1.51.Final</version>
+      <classifier>linux-x86_64</classifier>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -142,11 +142,19 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <!-- Test dependency is for Unix socket test -->
+      <!-- Test dependency is for Unix socket test (linux) -->
       <groupId>io.netty</groupId>
       <artifactId>netty-transport-native-epoll</artifactId>
       <version>4.1.51.Final</version>
       <classifier>linux-x86_64</classifier>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <!-- Test dependency is for Unix socket test (bsd/macos) -->
+      <groupId>io.netty</groupId>
+      <artifactId>netty-transport-native-kqueue</artifactId>
+      <version>4.1.51.Final</version>
+      <classifier>osx-x86_64</classifier>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -141,6 +141,14 @@
       <version>1.7.25</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
tsegismont, line 144: Can you also add the `kqueue` equivalent for those who want to build the project on Macs?

src/test/java/io/vertx/micrometer/UnixSocketTest.java
```diff
@@ -0,0 +1,91 @@
...
+        });
+    }
+
+    @Override
+    public void stop(Promise<Void> stopPromise) {
+      stopPromise.complete();
+    }
+  }
+
+  public class DomainSocketClientTriggerVertical extends AbstractVerticle {
```
tsegismont, line 70:
```suggestion
  public class DomainSocketClientTriggerVerticle extends AbstractVerticle {
```

Auto-correct? :wink:

src/test/java/io/vertx/micrometer/UnixSocketTest.java
```diff
@@ -0,0 +1,91 @@
...
+      .addDisabledMetricsCategory(MetricsDomain.EVENT_BUS)
+      .addLabels(Label.REMOTE)
+      .setRegistryName(registryName)
+      .setEnabled(true)))
+      .exceptionHandler(ctx.exceptionHandler());
+
+    Async allDeployed = ctx.async();
+    vertx.deployVerticle(
+      new DomainSocketServer(),
+      h -> vertx.deployVerticle(new DomainSocketClientTriggerVertical(), ch -> allDeployed.complete()));
```
tsegismont, line 41:
```suggestion
      h -> vertx.deployVerticle(new DomainSocketClientTriggerVerticle(), ch -> allDeployed.complete()));
```

tsegismont (approved):
@jotak thanks for the fast turnaround! Approved with some minor changes requested. Can you backport to 3.9 branch?

tsegismont (changes requested):
Actually it seems there's an issue with the test: actual connection is 0.

src/test/java/io/vertx/micrometer/UnixSocketTest.java
jotak: It was copy-pasted from the provided reproducer :-)

pom.xml
jotak: Can you give me some pointer to how I can do that?
Actually I don't have a mac so I wouldn't even be able to test, it would be better if someone who has one could take care of that? (winks @vietj )

jotak:
> Actually it seems there's an issue with the test: actual connection is 0.

It should be fixed : need to wait that connection gauge is decreased back to 0 after disconnection.

pom.xml
tsegismont: https://vertx.io/docs/vertx-core/java/#_native_bsd_transport

Don't test, we know it works :smile_cat:

jotak: So what does it mean? I just add this dependency and voilà, no code change?

jotak: (updated)

tsegismont: Yes, that's it. Netty will pick the right library at runtime.

tsegismont (approved):
Thanks @jotak !

jotak:
backported to 3.9

Issue #114, opened 2020-09-30
Rename some metrics to better match conventions


jotak:
Some metrics are a mix of snake case and camel case.
E.g. `vertx_http_client_responseTime_seconds`

It would be nice to take the opportunity of major release (vertx 4) to rename them.
As it's a breaking change for metrics consumers, some compatibility mode should be implemented.

PR #115, opened 2020-09-30
Rename some metrics to better match conventions

jotak:
Also:
- Update doc
- Update tests
- Add compatibility mode to stick with old names

Fixes https://github.com/vert-x3/vertx-micrometer-metrics/issues/114


src/main/java/io/vertx/micrometer/MicrometerMetricsOptions.java
```diff
@@ -387,4 +395,24 @@ public MicrometerMetricsOptions setJvmMetricsEnabled(boolean jvmMetricsEnabled)
...
+   * {@code MetricsNaming} is a structure that holds names of all metrics, each one can be changed individually.
+   * @return the configured {@code MetricsNaming} object (defaults to Vert.x names).
+   */
+  public MetricsNaming getMetricsNaming() {
+    return metricsNaming;
+  }
+
+  /**
+   * {@code MetricsNaming} is a structure that holds names of all metrics, each one can be changed individually.
+   * For instance, to retrieve compatibility with the names used in Vert.x 3.x, use {@code setMetricsNaming(MetricsNaming.v3Names())}
```
jotak, line 409: @vietj  don't quite remember the subtleties of the doc/code gen, is that "AT_code" tag within javadoc a valid thing?

src/main/asciidoc/index.adoc
```diff
@@ -407,12 +419,12 @@ depending on the backend used.
 |Gauge
 |Number of connections to the remote host currently opened.
 
-|`vertx_net_client_bytesReceived`
+|`vertx_net_client_read_bytes`
```
vietj, line 422: those are breaking changes ?

src/main/java/io/vertx/micrometer/MicrometerMetricsOptions.java
vietj: that's plain javadoc so it does not apply here

vietj: `{@code}` is fine (it's javadoc)

vietj:
it looks like a few metrics are renamed, e.g "bytes received". -> "read bytes" is that right ?

src/main/asciidoc/index.adoc
vietj: I think the preview wording was fine, i.e bytes_received

src/main/asciidoc/index.adoc
```diff
@@ -407,12 +419,12 @@ depending on the backend used.
...
 |`local`, `remote`
 |Counter
 |Number of bytes received from the remote host.
 
-|`vertx_net_client_bytesSent`
+|`vertx_net_client_written_bytes`
```
vietj, line 427: I think _bytes_written_ is better

src/main/asciidoc/index.adoc
```diff
@@ -453,22 +465,22 @@ depending on the backend used.
...
 |Time spent in queue before being processed.
 
-|`vertx_http_client_queue_size`
+|`vertx_http_client_queue_pending`
 |`local`, `remote`
 |Gauge
 |Number of pending elements in queue.
 
-|`vertx_http_client_requests`
+|`vertx_http_client_request_active`
```
vietj, line 478: _active_requests_

src/main/asciidoc/index.adoc
```diff
@@ -453,22 +465,22 @@ depending on the backend used.
...
 |`local`, `remote`, `path`, `method`
 |Gauge
 |Number of requests waiting for a response.
 
-|`vertx_http_client_requestCount`
+|`vertx_http_client_requests`
```
vietj, line 483: _total_requests_ ?

vietj:
a couple of suggestions on better v4 names

src/main/asciidoc/index.adoc
jotak: Depends ... do you prefer Prometheus conventions, openmetrics, or opentelemetry? :-)
Actually, I've been following the prometheus guidelines ( https://prometheus.io/docs/practices/naming/#metric-names ) but perhaps that's not such a good idea. Let me review all of this again.

src/main/asciidoc/index.adoc
jotak: Eventbus had bytesRead / bytesWritten, and net client bytesReceived, bytesSent. I'd like to make it more consistent. To me, the two are more or less equal, just have to choose one, and apply the snake-case naming.

src/main/asciidoc/index.adoc
jotak: If we follow prometheus conventions, that would be `vertx_http_client_requests_total`. Also note that "total" is automatically suffixed on counters by micrometer (at least with prometheus backends, I didn't check others) , which means if you write "total_requests" I think it will end up in "total_requests_total"

jotak: By the way, I realize I need to write in doc the final output that we'll see in prometheus, so in any case I have to change this.

src/main/asciidoc/index.adoc
vietj: oh sh*t :-)

does it make sense to have a naming class for each backend ? isn't micrometer able to do that as well ?

src/main/asciidoc/index.adoc
jotak: as discussed, I've changed to `xxx_bytes_read` / `xxx_bytes_written`.

This is a little infringement on prom's naming guidelines, which expect as pure convention to set unit at the end of the metric name, but on the flip side is closer to human language ; and anyway this prometheus guideline is made obsolete in opentelemetry/metrics if i'm correct.

src/main/asciidoc/index.adoc
jotak: done

src/main/asciidoc/index.adoc
jotak: changed to `vertx_http_client_requests_total` (`_total` being appended by micrometer for counters)

jotak:
@vietj I hope this is now OK

src/main/asciidoc/index.adoc
```diff
@@ -493,7 +505,7 @@ depending on the backend used.
 |Summary
 |Size in bytes of the responses.
 
-|`vertx_http_client_wsConnections`
+|`vertx_http_client_ws_connections`
```
vietj, line 508: how about _active_connections_ like _active_requests_ ?

src/main/asciidoc/index.adoc
```diff
@@ -509,21 +521,21 @@ depending on the backend used.
...
 |Counter
 |Number of bytes received by the Net Server.
 
-|`vertx_net_server_bytesSent`
+|`vertx_net_server_bytes_written`
 |`local`, `remote`
 |Counter
 |Number of bytes sent by the Net Server.
 
+|`vertx_net_server_connections`
```
vietj, line 534: how about _active_connections_ like _active_requests_ ?

src/main/asciidoc/index.adoc
```diff
@@ -540,37 +552,37 @@ depending on the backend used.
...
 |Counter
 |Number of bytes received by the HTTP Server.
 
-|`vertx_http_server_bytesSent`
+|`vertx_http_server_bytes_written`
 |`local`, `remote`
 |Counter
 |Number of bytes sent by the HTTP Server.
 
+|`vertx_http_server_connections`
```
vietj, line 565: how about _active_connections_ like _active_requests_ ?

src/main/asciidoc/index.adoc
```diff
@@ -540,37 +552,37 @@ depending on the backend used.
...
 |Number of requests being processed.
 
-|`vertx_http_server_requestCount`
+|`vertx_http_server_requests_total`
 |`local`, `remote`, `path`, `method`, `code`, `route`
 |Counter
 |Number of processed requests.
 
-|`vertx_http_server_requestResetCount`
+|`vertx_http_server_request_reset_total`
```
vietj, line 585: requests_reset ?

src/main/asciidoc/index.adoc
```diff
@@ -580,17 +592,17 @@ depending on the backend used.
...
-|Request processing time.
+|Request processing time in seconds.
 
 |`vertx_http_server_response_bytes`
 |`local`, `remote`, `path`, `method`, `code`, `route`
 |Summary
 |Size in bytes of the responses.
 
-|`vertx_http_client_wsConnections`
+|`vertx_http_client_ws_connections`
```
vietj, line 605: how about _active_connections_ like _active_requests_ ?

vietj:
there is a slight concern for gauges giving the number of active items.

on one side we do have _active_requests_ and all other use _connections_, I think either we should use the advjective _active_ everywhere or we should drop it on _requests_

src/main/asciidoc/index.adoc
jotak: yes, that makes sense

src/main/asciidoc/index.adoc
jotak: yes, I guess you meant `active_ws_connections`

src/main/asciidoc/index.adoc
jotak: `_total` is added automatically for counters for prometheus backend, which I take as a reference here for names

As for the plural on "requests" I'm unsure, "reset" here also is used as a noun so could be as well "requests_resets"? or "request_resets"?

src/main/asciidoc/index.adoc
jotak: done

src/main/asciidoc/index.adoc
jotak: done

src/main/asciidoc/index.adoc
jotak: done

src/main/asciidoc/index.adoc
jotak: Still not 100% sure of the most correct English form but "request resets" has more favors at the moment

PR #123, opened 2021-06-03
Fix broken test, missing "vertx.net.client.active.connections"

jotak:
When HTTP client requests are performed, now the metric "vertx.net.client.active.connections" is generated in addition to the expected "vertx.http.client.active.connections"

DNM until figured out what happens...

@vietj any idea what could trigger that change in the http client module ? To me it looks like a regression, I wouldn't expect metrics endpoint from the net client being invoked while using the http client.


vietj:
now the HttpClient is using the NetClient to establish connections, that might explain it.

vietj:
in vertx-core I could disable metrics of the NetClient used by HttpClient.

jotak:
> in vertx-core I could disable metrics of the NetClient used by HttpClient.

+1 if it's an easy one and makes sense to you

Issue #89, opened 2019-11-20
Custom Tag Provider


dtaveras:
The generation of tags is tied to Label types. Is there a way to add a custom tag?
I do understand that there is common tags in the MeterRegistry but i was more looking for one when creating the Meter. As an example I want to tag each meter with a user

jotak:
Hi @dtaveras ,

As you say, the MeterRegistry has common tags ; and you can get that registry to impact all registered metrics. E.g.

```java
    MeterRegistry registry = BackendRegistries.getDefaultNow();
    registry.config().commonTags("user", "alice");
```
Is that not what you're looking for?

dtaveras:
I was more interested in being able to extract more information based on the request, so having a tag provider with signature Request -> [Tag]. 
As example we also use https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/MetricsRequestEventListener.java which accepts a tag provider with the signature above.

dtaveras:
common tags will do for now, was just curious if i missed something.

jotak:
I think I get it, it seems like a good enhancement to add.

dtaveras:
Thanks, great job with this library. A ton of great features and easy to configure.

vietj:
@jotak do we have a contributor for this enhancement ?

jotak:
@vietj not that I know. I'm going to take a look.

PR #124, opened 2021-06-03
Implement custom tags provider for HTTP metrics

jotak:
- Allows to implement a CustomTagsProvider interface to provide tags out
of HttpRequest objects.
- Add tests

Fixes https://github.com/vert-x3/vertx-micrometer-metrics/issues/89

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -369,6 +369,19 @@ Here is an example to replace the actual `path` label of HTTP requests with a ge
 
 NOTE: Matchers use MeterFilters under the hood.
 
+=== Custom tags provider
+
+You can define a function that generates additional tags (or labels) for HTTP server metrics.
+Such function takes an {@link io.vertx.core.spi.observability.HttpRequest} object as a parameter, and returns
+an Iterable of {@link io.micrometer.core.instrument.Tag}.
+
+As an example, here is how to map the _x-user_ header to a custom label _user_:
+
+[source,$lang]
+----
+{@link examples.MicrometerMetricsExamples#useCustomTagsProvider()}
+----
+
 === Snapshots
 
 A {@link io.vertx.micrometer.MetricsService} can be created out of a {@link io.vertx.core.metrics.Measured} object
--- a/src/main/java/io/vertx/micrometer/MicrometerMetricsOptions.java
+++ b/src/main/java/io/vertx/micrometer/MicrometerMetricsOptions.java
@@ -16,14 +16,17 @@
 package io.vertx.micrometer;
 
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.metrics.MetricsOptions;
 import io.vertx.core.spi.VertxMetricsFactory;
+import io.vertx.core.spi.observability.HttpRequest;
 
 import java.util.*;
+import java.util.function.Function;
 
 /**
  * Vert.x micrometer configuration.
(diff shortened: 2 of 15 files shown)


jotak:
Hey @vietj  perhaps it needs a little bit of discussion prior to merging.
To summarize, the feature is about having the ability to customize metric tags out of request data, in particular headers. Basically, tags can be injected for every request based on some request or headers mapping function.

Here I've implemented the most straightforward way, which is by having the user providing a `CustomTagsProvider` mapper into config, something like that:

```java
    vertx = Vertx.vertx(new VertxOptions().setMetricsOptions(new MicrometerMetricsOptions()
      .setRequestsTagsProvider(req -> {
        String user = req.headers().get("user");
        return Collections.singletonList(Tag.of("user", user));
      })
      .setPrometheusOptions(new VertxPrometheusOptions().setEnabled(true))
      .setEnabled(true)));
```

* Pros: it's flexible; The user can create tags the way she wants out of an `HttpRequest` object
* Cons: non-serializable, the `setRequestsTagsProvider` option is executable code, cannot be set up via json

So, we can imagine an alternative where a similar feature would be achieved by providing a (serializable) option, e.g. a list of header keys or a list of pairs [header key, tag name] that would be used for a mapping of headers to tags. But it would be less flexible, no possibility of combining headers, or using other info from `HttpRequest`, or having more complex behaviours. So people may feel limited.

WDYT? Are we good with my first idea; and eventually we can iterate later to enhance the feature by adding serializable options if that's what users want?

jotak:
PS: CI failure is related to https://github.com/vert-x3/vertx-micrometer-metrics/pull/123

narzach:
@jotak This is great, thank you for taking the time to make this happen!

IMO your approach is exactly what we need. The serializable option would never be as useful as this is, and IMO would not provide enough value to be worth doing. This gives us full control to set tags based on the context request. For my use cases, I need the ability to do exactly what you're doing, extracting values from HTTP path parameters, query string, or headers, and then injecting corresponding tags. Static tags that are added per app will be useful, yes, but that is trivial and can be done with your setRequestsTagsProvider() approach.

Curious if anyone else has had a chance to look at this?

src/main/java/io/vertx/micrometer/impl/CustomTagsProvider.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package io.vertx.micrometer.impl;
+
+import io.micrometer.core.instrument.Tag;
+
+/**
+ * @author Joel Takvorian
+ */
+public interface CustomTagsProvider<T> {
```
vietj, line 24: this interface is in an impl package, I think instead we should rely on Java function types to do this, i.e the provider should simply be a `Function<T, Iterable<Tag>>`

src/main/java/io/vertx/micrometer/MicrometerMetricsOptions.java
```diff
@@ -66,6 +68,7 @@
   private VertxJmxMetricsOptions jmxMetricsOptions;
   private boolean jvmMetricsEnabled;
   private MetricsNaming metricsNaming;
+  private CustomTagsProvider<HttpRequest> requestsTagsProvider;
```
vietj, line 71: instead use `Function<HttpRequest, Iterable<Tag>>`

vietj (changes requested):
use a function instead of a specific new type

jotak:
@vietj done

vietj (changes requested):
Documentation

src/main/asciidoc/index.adoc
```diff
@@ -369,6 +369,19 @@ Here is an example to replace the actual `path` label of HTTP requests with a ge
 
 NOTE: Matchers use MeterFilters under the hood.
 
+=== Custom tags provider
+
+It is possible to define a function that generates additional tags (or labels) for HTTP server metrics.
```
vietj, line 374: _You can define_

src/main/asciidoc/index.adoc
```diff
@@ -369,6 +369,19 @@ Here is an example to replace the actual `path` label of HTTP requests with a ge
...
+This function takes an {@link io.vertx.core.spi.observability.HttpRequest} object as a parameter, and returns
```
vietj, line 375: Such function

jotak: done

src/main/asciidoc/index.adoc
jotak: done

PR #132, opened 2022-01-26
Bump micrometer from 1.6.2  to 1.8.4

yeikel:
Motivation:

Maintenance/ Keep dependencies up to date

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,7 @@
     <assertj.version>3.9.0</assertj.version>
     <junit.version>4.13.1</junit.version>
     <maven-failsafe-plugin.version>2.18.1</maven-failsafe-plugin.version>
-    <micrometer.version>1.6.2</micrometer.version>
+    <micrometer.version>1.8.4</micrometer.version>
     <jar.manifest>${project.basedir}/src/main/resources/META-INF/MANIFEST.MF</jar.manifest>
   </properties>
 


tsegismont:
@yeikel on the vertx user group you said you add issues with metrics when upgrading to this version but it seems our test suite does not detect any regression.

Can you elaborate about the issues?

yeikel:
> @yeikel on the vertx user group you said you add issues with metrics when upgrading to this version but it seems our test suite does not detect any regression.
> 
> Can you elaborate on the issues?




@tsegismont 

Are you referring to
 https://groups.google.com/g/vertx/c/9mCsl4rKhhc/m/Jaeh64rPAQAJ?


I am not the one who reported problems with this version. The email you're referring to was signed by "Clive"

tsegismont:
@yeikel yes I was. Sorry for the confusion.

yeikel:
> @yeikel yes I was. Sorry for the confusion.

No problem! Thank you for checking

Let me know if you publish any snapshot for this upgrade so I can also test it as well

tsegismont:
I'm waiting for a reply from the user. Besides, this would be a Vert.x
4.3.0 upgrade so I can't merge yet (master branches still target 4.2).

Anyway, given this is a simple POM version, you could test right away by
changing the micrometer dependencies in your project's POM file.

yeikel:
Please note that `1.8.3` was just released and I just updated this PR instead

yeikel:
Note that `1.8.4` was just released and I updated the PR instead

tsegismont (approved):
Thank you @yeikel

PR #138, opened 2022-03-20
add dependabot configuration

yeikel:
See https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
yeikel, line 6: If needed, we can change this to weekly

tsegismont:
Thank you but I don't believe this is necessary. GitHub provides notifications about releases that we use to determine if a new version is required.

yeikel:
> Thank you but I don't believe this is necessary. GitHub provides notifications about releases that we use to determine if a new version is required.

Are these pull requests created automatically?

Because notifications are easy to miss and it's not only about security alerts

tsegismont:
No they are not created automatically.

PR #151, opened 2022-07-07
allow NoopGauges. fixes #150

### LINKED ISSUE #150: Using Micrometer Filters to disable any VertX Gauge measurement causes infinite loop
### Version

VertX 4.3.2

### Context

(https://github.com/vert-x3/vertx-micrometer-metrics/commit/5a4a10a16695b8d81866d0325e9125bc63e8d43d)

In our microservices we are filtering out certain metrics from micrometer / prometheus to reduce the overall load (we dont need them) and this also includes specific metrics tracked by vertx like:

```
"vertx.pool.ratio"
"vertx.pool.completed"
"vertx.eventbus.handlers"
```

for now i fixed this for us by changing from using Micrometer Filter's to disable the metrics to whitelisting the metrics to be scraped instead but it would be nice if the `NullGauge` from Micrometer would be supported.

```
final var whitelist = pmr.getMeters().stream()
              .map(Meter::getId)
              .filter(id -> !filteredMetrics.contains(id.getName()))
              .map(id -> id.getConventionName(NAMING_CONVENTION))
              .collect(Collectors.toUnmodifiableSet());
            System.err.println(whitelist);
            return ((PrometheusMeterRegistry) pmr).scrape(TextFormat.CONTENT_TYPE_004, whitelist);
```

### Do you have a reproducer?

no - can provide one on demand

### Steps to reproduce

```
final var reg = BackendRegistries.getDefaultNow();
 reg.config().meterFilter(MeterFilter.deny(id -> "vertx.eventbus.handlers" == id.getName())); // we check for a whole configurable list here
// filters do not retroactively remove metrics which have been collected already - has to be done afterwards manually once
reg.forEachMeter(meter -> {
  if ("vertx.eventbus.handlers" == id.getName())) {
    reg.remove(meter.getId());
  }
});
```

### Extra

openJdk 18

DemonicTutor:
Motivation:

https://github.com/vert-x3/vertx-micrometer-metrics/issues/150

maybe i am missing something obvious but i think this would be a viable approach to implement the gauges.

from VertX point of view the objects (LongAddr) are always there and used - handed out thread save and cannot be GC'd.
from Micrometer point of view - it can internally either collect the objects or not if they are removed later.

Conformance:

### DIFF
--- a/src/main/java/io/vertx/micrometer/impl/meters/Gauges.java
+++ b/src/main/java/io/vertx/micrometer/impl/meters/Gauges.java
@@ -18,8 +18,6 @@
 package io.vertx.micrometer.impl.meters;
 
 import io.micrometer.core.instrument.*;
-import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
-import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import io.vertx.micrometer.Label;
 import io.vertx.micrometer.impl.Labels;
 
@@ -37,7 +35,7 @@ public class Gauges<T> {
   private final Supplier<T> tSupplier;
   private final ToDoubleFunction<T> dGetter;
   private final MeterRegistry registry;
-  private final ConcurrentMap<Meter.Id, Object> gauges;
+  private final ConcurrentMap<Meter.Id, Object> candidatesByMeterId;
 
   public Gauges(ConcurrentMap<Meter.Id, Object> gauges,
                 String name,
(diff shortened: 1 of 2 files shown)


tsegismont:
Thanks for the PR. The issue has been fixed in #152.

The issue with the proposed changes is that you have no guarantee that, when the lookup method is invoked concurrently, the value store in the gauges map is actually the value referenced by the gauge (only one gauge will be registered but we do not know in advance which one).

PR #170, opened 2023-02-27
fix(noopGauge): fix NoopGauge going to infinite

cristian-joseb:
Motivation:

Explain here the context, and why you're making that change, what is the problem you're trying to solve.

Conformance:
When applying a Meter filter that return MeterFilterReply.DENY, a Gauge of type NoopGauge is returned, which is making the function not being executed then gauge.value() is called. This cause the Gauges.get method going on a infinity loop because of the `if (candidateFunc.invoked)` always returning false.


cristian-joseb:
@tsegismont when is this planned to be applied, we want to update out code base, since we still are working on migrating to Vertx4.

cristian-joseb:
I saw the PR was merged, thank you @tsegismont :)

tsegismont:
You're welcome

Issue #184, opened 2023-05-24
Performance analysis and improvements


vietj:
- Performance analysis on typical use case
- Measure overhead (GC, CPU, throughput, latency)
- Compare overhead to other implementations (Quarkus, Spring)
- Make changes for new target
- Possible breaking API and metric names changes
- Evaluate limitations due to Micrometer implementation

tsegismont:
Comparison of throughput using Techempower benchmark Vert.x core implementation

![image](https://github.com/vert-x3/vertx-micrometer-metrics/assets/1500598/9f3380f1-4323-4233-a807-37e46aca2fb7)

Up to 64% reduced throughput when Micrometer Metrics is enabled.

CPU profiling shows a lot of time is spent on reporting metrics:

![image](https://github.com/vert-x3/vertx-micrometer-metrics/assets/1500598/73d8a6f8-7e37-4350-9740-28bdcd83d6d1)

Digging a little deeper, dynamic resolution of the meters is actually the problem:

![image](https://github.com/vert-x3/vertx-micrometer-metrics/assets/1500598/0df3de26-fd8b-41ac-a3e7-230440bcfe0c)

You can see this by the calls to the following methods:

- `io.vertx.micrometer.impl.meters.Counters#get(java.lang.Iterable<io.micrometer.core.instrument.Tag>, java.lang.String...)`
- `io.vertx.micrometer.impl.meters.Gauges#get(java.lang.Iterable<io.micrometer.core.instrument.Tag>, java.lang.String...)`
- ...

tsegismont:
Comparison of throughput using Techempower fortunes benchmark (Vert.x core implementation)

![image](https://github.com/vert-x3/vertx-micrometer-metrics/assets/1500598/2bceee00-af11-4125-aeaa-f0973fbcd85d)

CPU profiling, with metrics related frames in fuchsia (Vert.x + Micrometer)

![image](https://github.com/vert-x3/vertx-micrometer-metrics/assets/1500598/354aa5d4-2410-4543-814a-46c6f58fc082)


Up to 10% reduced throughput when Micrometer Metrics is enabled.

tsegismont:
With the following changes:

- Avoid meters lookup where possible  21d9f758c03704951bc672b57278bfd1340a0cba
- Filter tags up front instead of using a Micrometer meter filter a4e266b9365e1e1c6d5e4d8cad2d0691a9d374b6
[...]

PR #200, opened 2023-10-11
Performance improvements

tsegismont:
Each commit can be reviewed individually.

Motivation for changes and measured improvements can be found in https://github.com/vert-x3/vertx-micrometer-metrics/issues/184


src/main/java/io/vertx/micrometer/MetricsNaming.java
```diff
@@ -576,4 +576,51 @@ public MetricsNaming setPoolQueuePending(String poolQueuePending) {
     this.poolQueuePending = poolQueuePending;
     return this;
   }
+
+  public MetricsNaming withBaseName(String baseName) {
```
vietj, line 580: is that called often ? if yes we should consider using a stringbuilder with basename and reuse it for each string we want to produce

src/main/java/io/vertx/micrometer/impl/AbstractMetrics.java
```diff
@@ -70,23 +80,59 @@ public String baseName() {
     return category == null ? null : "vertx." + category + ".";
   }
 
-  Counters counters(String name, String description, Label... keys) {
-    return new Counters(baseName() + name, description, registry, keys);
+  Counter counter(String name, String description, Tags tags) {
+    Meter.Id id = new Meter.Id(name, tags, null, description, COUNTER);
+    Counter c = meterCache.get(id);
```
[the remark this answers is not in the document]
vietj: I see the impl is currently using a thread local, but if we make this pluggable for the user, we should have a CAS operation

src/main/java/io/vertx/micrometer/impl/VertxEventBusMetrics.java
```diff
@@ -15,152 +15,152 @@
...
-  private static boolean isValid(Handler handler) {
-    return handler != null && handler.address != null;
-  }
-
-  static class Handler {
-    private final String address;
-
-    Handler(String address) {
-      this.address = address;
+  class Handler {
```
vietj, line 144: Handlers ?

src/main/java/io/vertx/micrometer/MetricsNaming.java
tsegismont: No, it's called when a metrics object is created (i.e. when Vert.x invokes methods like `io.vertx.core.spi.metrics.VertxMetrics#createHttpServerMetrics`

src/main/java/io/vertx/micrometer/impl/AbstractMetrics.java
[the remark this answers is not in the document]
tsegismont: Right, `MeterCache` is thread-safe because it's implemented with a map stored `FastThreadLocal`.

It is not planned to make this pluggable.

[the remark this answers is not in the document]
vietj: ok then I think we should simply remove the abstraction and not have a cache that ignores everything and instead check nullity that would simplify the code

src/main/java/io/vertx/micrometer/impl/VertxEventBusMetrics.java
tsegismont: I didn't change that name. For consistency, I can rename it to `HandlerMetric` (we have `NetServerSocketMetric`, `RequestMetric`, ... etc)

tsegismont: Done, see bfce025

src/main/java/io/vertx/micrometer/impl/AbstractMetrics.java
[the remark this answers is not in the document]
tsegismont: Done

src/main/java/io/vertx/micrometer/MetricsNaming.java
vietj: ah right, that's fine then!

src/main/java/io/vertx/micrometer/impl/MeterCache.java
```diff
@@ -0,0 +1,56 @@
...
+
+package io.vertx.micrometer.impl;
+
+import io.micrometer.core.instrument.Meter;
+import io.netty.util.concurrent.FastThreadLocal;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class MeterCache {
```
vietj, line 26: make it package private

src/main/java/io/vertx/micrometer/impl/VertxEventBusMetrics.java
vietj: ah right, I thought it was named Handler because it was grouping of bunch of field named handlers, that makes sense then

src/main/java/io/vertx/micrometer/impl/MeterCache.java
tsegismont: Done

src/main/java/io/vertx/micrometer/impl/MeterCacheImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+import io.micrometer.core.instrument.Meter;
+import io.netty.util.concurrent.FastThreadLocal;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class MeterCacheImpl implements MeterCache {
+
+  private final FastThreadLocal<Map<Meter.Id, Object>> threadLocal;
```
franz1981: @tsegismont we recently had a problem with @vietj on the CombinerExecutor because of a non static Fast thread local, which was causing some silly leak (which could lead to CVE) due to how it works under the hood.
TLDR make sure not many cache instances could be created in the lifecycle of the application!

tsegismont: In fact, we talked about this @vietj already :smile:

I replaced `FastThreadLocal` cache with a (less efficient) Vert.x EL Context cache. And then Micrometer team told us about their `MeterProvider` API, which gave good enough results (see the end of the blog post published yesterday).

Anyway, thanks for the heads-up!

Issue #213, opened 2024-03-03
Add support for customizing Prometheus request for metrics scraping by introducing PrometheusRequestHandler


swamymavuri:
#### Describe the feature
This feature aims to enhance the customization options for Prometheus request handling by introducing the PrometheusRequestHandler. This handler provides users with the ability to define their own request handling strategies tailored to their application's needs, enabling greater flexibility and control over how Prometheus metrics are scraped and served.

#### Proposal:
The proposal involves adding support for customizing Prometheus request handling through the introduction of the PrometheusRequestHandler. This handler can be configured to scrape metrics from a PrometheusMeterRegistry and serve them over HTTP, offering various configuration options for metrics endpoints and the Prometheus registry.
#### Use cases

  **Enhanced Flexibility**: Users can define custom request handling procedures tailored to their application's unique requirements by customizing registry.

  **Adaptability**: By allowing users to define their own request handling logic, the feature promotes adaptability to diverse project environments and use cases.

#### Implementation Details
The implementation involves:
- Introducing the PrometheusRequestHandler interface with factory methods for creating handlers.
- Providing configuration options for specifying PrometheusMeterRegistry and metrics endpoints.
- Implementing the PrometheusRequestHandlerImpl class to handle HTTP requests for Prometheus metrics scraping.

`
 PrometheusMeterRegistry customRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
 customRegistry.config().namingConvention(NamingConvention.camelCase);
router.route("/custom-metrics").handler(ctx -> PrometheusRequestHandler.create(customRegistry,"/custom-metrics")
      .handle(ctx.request()));
`

This feature enhancement will enrich the Prometheus integration in Vert.x, offering users greater flexibility and control over how metrics are handled and served.

#### Contribution

Who should implement this feature ? are you volunteering for implementing this feature - **YES**

PR #214, opened 2024-03-03
Added support for custom exception handler for VertxPrometheusOptions…

swamymavuri:
… (#213)

Motivation:

Explain here the context, and why you're making that change, what is the problem you're trying to solve.

Conformance:


tsegismont (commented):
Thanks for your proposal. Instead of adding an exception handler to options, it would be better to create a class that provides a request handler (similar to the scraping handler for vert.x web).
This way, the user can create the HTTP server without implementing the request handler and having full control on the server behavior.

swamymavuri:
Hi @tsegismont ,
Could you please review the following code changes as per your suggestions

Created a PrometheusRequestHandler class to manage Prometheus requests.
By default, PrometheusRequestHandler  is used while creating HttpServer if embeddedServer is enabled.

and User can get the flexibility to create their own httpServer without implementing the request handler

Thanks

swamymavuri:
> Can you please sign the Eclipse Contributor Agreement and then, using the same Git author/user email, sign-off the commit?

Thanks @tsegismont , revalidated the commit by signing off the agreement.

src/main/java/io/vertx/micrometer/PrometheusRequestHandler.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * A request handler for serving Prometheus metrics.
+ *
+ * @author Swamy Mavuri
+ */
+public class PrometheusRequestHandler implements Handler<HttpServerRequest> {
+
+  private final PrometheusMeterRegistry registry;
+  private final VertxPrometheusOptions options;
```
tsegismont, line 17: Can you change this to a `String` field that stores the embedded server endpoint? We don't need all of the options.

src/main/java/io/vertx/micrometer/PrometheusRequestHandler.java
```diff
@@ -0,0 +1,40 @@
...
+   */
+  public PrometheusRequestHandler(PrometheusMeterRegistry registry, VertxPrometheusOptions options) {
+    this.registry = registry;
+    this.options = options;
+  }
+
+
+  @Override
+  public void handle(HttpServerRequest request) {
+    if (options.getEmbeddedServerEndpoint().equals(request.path())) {
```
tsegismont, line 33: Here you can simplify the condition to use the string field

src/main/java/io/vertx/micrometer/PrometheusRequestHandler.java
```diff
@@ -0,0 +1,40 @@
...
+import io.vertx.core.Handler;
+import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.http.HttpServerRequest;
+
+/**
+ * A request handler for serving Prometheus metrics.
+ *
+ * @author Swamy Mavuri
+ */
+public class PrometheusRequestHandler implements Handler<HttpServerRequest> {
```
tsegismont, line 14: Can you please create an interface annotated with `@VertxGen` like the `PrometheusScrapingHandler`

```
@VertxGen
public interface PrometheusScrapingHandler {

  /**
   * Creates a Vert.x Web {@link io.vertx.ext.web.Route} handler for Prometheus metrics scraping.
   * The default backend registry is used.
   *
   * @return a {@link io.vertx.ext.web.Route} handler for the default backend registry
   * @see BackendRegistries#getDefaultNow()
   */
  static Handler<RoutingContext> create() {
    return new PrometheusScrapingHandlerImpl();
  }
```

If you do this, we will automatically have generated files for RxJava, Mutiny, ...etc

tsegismont (commented):
Thanks for the updates @swamymavuri

tsegismont:
Can you please also change the PR and issue titles and descriptions to reflect the new direction?

PR #218, opened 2024-04-10
Remove cache and use MeterProvider for dynamic tags

tsegismont:
Micrometer will provide an internal cache in v1.13.

Besides, since v1.12, to avoid creating builders again and again, we can use the MeterProvider API. MeterProvider allows, using a unique builder, to change tags dynamically.


tsegismont:
See https://github.com/micrometer-metrics/micrometer/issues/4856#issuecomment-2042299278 for measurements

PR #241, opened 2024-10-22
Bump micrometer version to 1.13.6

pjgg:
Motivation:

Update 1-year-old library. Currently, in my company project, we have a monorepo with a single dependency version. Due to some other libraries, we need to use the latest Micrometer version. The main problem is that Vert.x 4.5.10 is still pointing to `io.micrometer.prometheus` instead of` io.micrometer.prometheusmetrics`, so I am getting the following error:

`error: package io.micrometer.prometheus does not exist`

Conformance:


pjgg:
Well, maybe this PR introduces a "breaking change" on Vertx 4.4.x isn´t it? because we are changing the micrometer package name ... so could not be introduced in this version. WDYT @tsegismont ?

vietj:
make it only in vertx 5 please

tsegismont:
@vietj Vert.x 5 already depends on 1.13.6, so nothing to do there

@pjgg yes, this would be a breaking change in Vert.x 4

Issue #262, opened 2025-06-17
[JPMS] Backend registries required even if not used


himanshumps:
### Version

5.0.0

### Context

Do we really need these ? This is breaking a pure JPMS based application and I have to include micrometer graphite, influx and .. even though they are not needed

https://github.com/vert-x3/vertx-micrometer-metrics/blob/d99644e6b3e936481ae3971f2831c88cefc33726/src/main/java/module-info.java#L13-L16

https://github.com/vert-x3/vertx-micrometer-metrics/blob/d99644e6b3e936481ae3971f2831c88cefc33726/src/main/java/module-info.java#L8

### Steps to reproduce

NA

himanshumps:
Google group link: https://groups.google.com/g/vertx/c/Zk-p1mt5w34

himanshumps:
Hi @vietj 

The modules are mentioned as required and once used in pure JPMS application, because influx, graphite and others are required, we need to include their dependencies even though it might be getting used in application

tsegismont:
Fixed by https://github.com/vert-x3/vertx-micrometer-metrics/pull/264 (not merged yet, waiting for build issues to be resolved)

PR #264, opened 2025-06-17
[JPMS] Backend registries should only be required for compilation

tsegismont:
See #262

### DIFF
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -5,15 +5,17 @@ module io.vertx.metrics.micrometer {
   requires static io.vertx.codegen.json;
 
   requires io.netty.buffer;
-  requires io.prometheus.metrics.model;
   requires io.vertx.core;
   requires io.vertx.core.logging;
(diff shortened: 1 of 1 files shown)


src/main/java/module-info.java
```diff
@@ -5,15 +5,17 @@
...
   requires io.vertx.core.logging;
   requires io.vertx.web;
   requires micrometer.core;
-  requires micrometer.registry.graphite;
-  requires micrometer.registry.influx;
-  requires micrometer.registry.jmx;
-  requires micrometer.registry.prometheus;
+
+  // Required only at compilation (users can pick the backends they want)
+  requires static micrometer.registry.graphite;
```
vietj, line 14: I think we should document this in https://vertx.io/docs/guides/modular-vertx-guide/#_micrometer_metrics ?

tsegismont: Will do

vietj: thanks a lot!

PR #266, opened 2025-07-14
Bump micrometer to 1.15.2

yeikel:
https://github.com/micrometer-metrics/micrometer/releases/tag/v1.15.2


yeikel:
> @yeikel thanks. Could you please use an email address verified with ECA to author the commit?

it seems that it expired. I signed it again and it should be valid now 

image

Are you able to retrigger the check?

Issue #267, opened 2025-07-16
Enable dependabot updates


yeikel:
### Describe the feature

Hi team, 

As usual, I am happy to help and raise maintenance  pull requests such as https://github.com/vert-x3/vertx-micrometer-metrics/pull/266

However, tools like dependabot can help us with this work. 

Would you be open to enable dependabot for this project? (and other Vertx projects in the future) 

We can start small with a limited list of dependencies or types (minor, patch) and go from there

### Contribution

I (@yeikel) would be happy to help

tsegismont:
There are a few Vert.x projects that have configured dependabot. It could be useful to let it upgrade Micrometer.

Can we configure it to:

- watch only Micrometer upgrades
- propose PRs for micro releases only in `4.x` and `5.0` branches
- propose PRs for all releases in the main branch.

Unrelated to this repo: would be nice if we could do the same for GraphQL-Java

vietj:
@tsegismont I think we should stick with updates that are not shared cross stack updates in the main branch

tsegismont:
@vietj that was the [idea](https://github.com/vert-x3/vertx-micrometer-metrics/issues/267#issuecomment-3084204632), updates for Micrometer only

Or do I misunderstand your comment?

PR #268, opened 2025-07-20
Add dependabot configuration to keep micrometer up to date

yeikel:
As discussed in https://github.com/vert-x3/vertx-micrometer-metrics/issues/267


.github/dependabot.yml
```diff
@@ -0,0 +1,41 @@
+# Dependabot configuration to keep micrometer up to date
+# The requirements were defined in https://github.com/vert-x3/vertx-micrometer-metrics/issues/267
+version: 2
+updates:
+  # Propose PRs for all the release types in the master branch.
```
vietj, line 5: we need to not allow this out of the box, because for main branch we should not allow a new major that has breaking changes

vietj: I mean I think we can allow but we need to check that micrometer does not bring breaking API changes when the upgrade happens

tsegismont: That's the plan, yes. I believe dependabot links to releases pages when they exist and also computes a list of changes when possible

yeikel: Yes, it will include releases notes

Also, our CI may also flag these too

tsegismont: CI will find breaking changes in the meters API or Prometheus registry implementation, but we don't have tests for every registry.

Anyway, in all cases we need to review the release notes before upgrading.

yeikel:
Thanks for the feedback. Is there anything else I should consider before moving this one forward?

tsegismont:
> Thanks for the feedback. Is there anything else I should consider before moving this one forward?

I have a question about config but that's it for now

